PR #7, opened 2018-03-27
Replaces $var with $_GET['var']

Watilin:
Assuming GET variables are automatically expanded in the global space, is assuming the `register_globals` config is on. This presents a security risk and has been deprecated since PHP 5.3.0, see http://php.net/manual/en/security.globals.php
This pull request updates all localizations of the document “htaccess url rewriting using mod_rewrite” to reflect PHP’s current best practices.

### DIFF
--- a/pages/web/hosting/htaccess_url_rewriting_using_mod_rewrite/guide.en-gb.md
+++ b/pages/web/hosting/htaccess_url_rewriting_using_mod_rewrite/guide.en-gb.md
@@ -89,7 +89,7 @@ These rules run the testing.php script with the GET variable containing the URL.
 ```
 <?
 print("testing server <br/>\n");
-print("var: $var\n");
+print("var: {$_GET['var']}\n");
 ?>
 ```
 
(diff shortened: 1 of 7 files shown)


VincentCasse:
Oh my god! Nobody codes PHP like that from 2005.

Good for me !
@Fracasx you can merge it :)

theodoreb:
If security is a concern we should avoid code vulnerable to XSS attacks too. 

The following is safe: 
```php
<?php
print "testing server <br/>\nvar: ";
print htmlspecialchars($_GET['var'], ENT_QUOTES);
```

Watilin:
@theodoreb you’re totally right and I didn’t think about that.
In the context of a tutorial-type document, it may not be a good thing to sacrifice code lisibility for maximum security.
On the other hand, OVH guides address real world scenarios and are not made for beginners.
Not sure what to think here… Your opinions, others?

Fracasx:
@Watilin  honestly I have mixed feelings to be honest. Security is really important for us all, and adding just a bit more code to improve it may be relevant.

@VincentCasse your opinion on it? (or others of course).

PR #21, opened 2018-04-18
Updated information to 7700k

mchanovh:
MC-64-OC now uses a 7700K instead of 6700K.
Will need to add new test results and screenshots to show the new CPU frequency.


Fracasx:
Changes validated, tech info OK : https://www.intel.com/content/www/us/en/products/processors/core/i7-processors/i7-7700k.html

PR #23, opened 2018-04-18
Fixed line break

mchanovh:
Changed links to be relative

### DIFF
--- a/pages/cloud/vps/secure_your_vps/guide.en-asia.md
+++ b/pages/cloud/vps/secure_your_vps/guide.en-asia.md
@@ -65,13 +65,14 @@ nano /etc/ssh/sshd_config
 > The command `nano` is given as an example; you can use the `vim` command, or any other command that allows you to edit the sshd_config file.
 >
 
-You then need to visualise the following line:
+You then need to find the following line:
 
 ```sh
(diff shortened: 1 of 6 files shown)


pages/cloud/vps/secure_your_vps/guide.en-asia.md
```diff
@@ -193,7 +194,7 @@ It is recommended that you create and adjust rules on this firewall, to fit your
 
 OVH solutions include a firewall at the entrance to the infrastructure, called the Firewall Network. Its implementation and configuration allow protocols to be blocked before they even arrive on your server.
 
-We also have a guide on [configuring this firewall network](https://docs.ovh.com/asia/en/dedicated/firewall-network/){.external}.
+We also have a guide on [configuring this firewall network](../../dedicated/firewall-network/){.ref}.
```
Fracasx: No relative, on purpose. Only full URL with {.external}, no {.ref}

Fracasx (changes requested):
All relative URL must be let this way, was put like it on purpose. Once changed PR will be validated.

mchanovh:
It wasn't consistent before, some of the links didn't have {.external} at all, why not use relative links to keep them the same between languages?

PR #29, opened 2018-04-30
Network configuration for Debian 9/Ubuntu 18

serverok:
route command is deprecated. Need to use "ip route" command instead of "route"


Fracasx:
Just a small check on one point before validation.

PR #42, opened 2018-05-23
Update universe.de-de.md

Joshua2504:
More german-friendly


Fracasx:
Same, I vadidate quickly ;) Just need a validation before

PR #44, opened 2018-05-23
Update product.de-de.md

Joshua2504:


Fracasx:
Hi @Joshua2504 , I just need validation from our german edition team. I validate (or amend) quickly.

Fracasx:
Current excerpt is correct this way, no particular need to put "bei OVH".

PR #53, opened 2018-06-26
Create guide.it-it.md

aport1996:
Translation of the FR guide in Italian.


Fracasx:
Hi @Nyxbiker , guide must be totally recreated (with updates based on Guide style + editorial work on it). We'll update is quickly (FR + all others versions, including IT).

PR #61, opened 2018-07-31
FR update ajouter une ressource a l heure

AsLY4:
Update with new template and screen capture from new plugin


AsLY4:
replaced by #63

PR #66, opened 2018-08-07
Update kernel url

gaotan:
OVH Kernels are now plublished here : 

https://last-public-ovh-kernel.snap.mirrors.ovh.net


Fracasx:
Hi @gaotan ,

All docs have been updated online :) https://docs.ovh.com/gb/en/dedicated/updating-kernel-dedicated-server/

Thanks for your help.

PR #69, opened 2018-08-08
re-order steps in CentOS session

mr-ssd:
it make sense to update after added repository


Fracasx:
Hi @sida-khalibre ,

I'll check for the global update of the whole guide according to guide style and proper tech info. I hold this request for the moment, time for me to check for the global update (to avoid fixes line by line). I will keep you informed soon.

PR #74, opened 2018-08-13
FR update vmware_drs_distributed_ressource_scheduler_new

AsLY4:
Fix slug and section for this guide prior to another PR


Fracasx:
Section fixed but not the slug (I already have traffic, I don't change slug even with redirection).

PR #84, opened 2018-09-23
Update guide.fr-fr.md

edouardberthe:
Adds details in case of a needed update of e2fsck.


BaptisteOVH:
Thanks for contributing! We updated the guide on our side with some modification.

PR #116, opened 2018-11-12
howto create VM image with packer

ticapix:


OVH-Team-Guides:
Merge done, thanks for contributing.

PR #120, opened 2018-11-19
data compute docs updated. There is no need to token anymore. Users n…

mojtabaimani:
Users no need to generate token anymore. They just need to download their openrc file. Also mac version of CLI binary is available.


OVH-Team-Guides:
Merge done, thanks for your contribution.

PR #147, opened 2019-02-05
Some words about X-OVH-TO-FREEZE

pdepaepe:
Add to Logs Data Platform convention guide, some words about the possibility to generate an other archive with user-defined lines only.

Signed-off-by: Pierre De Paepe <pierre.de-paepe@corp.ovh.com>


OVH-Team-Guides:
Merge done, thanks for your contribution.

PR #148, opened 2019-02-14
[webhosting] Update output address for cluster017

rclsilver:
After migration of cluster017 from P19 to GRA, output address has changed.


OVH-Team-Guides:
Merge done, thanks for your contribution

PR #178, opened 2019-03-22
EN update picture on stp psrd change

benoitest:


OVH-Team-Guides:
Was a test, will not be merged.

PR #185, opened 2019-04-05
secure vps, network firewall, rescue and ipalias

talmortis:
minor changes to **secure vps** and **network firewall** guides

---
### Ubuntu
while working on this guide mostly minor changes, except moving the Ubuntu part to the debian-like systems.
**this part does not exist in the French guide yet!**

### Solaris
**I'm not sure if we still offer Solaris, where this guide works.** 
My research showed a change of networking in Solaris 11.
We currently offer smartOS but that's mutation of Solaris11, so the config doesn't really work.

---
## rescue_mode
The screenshot **04** still showed a server name -> replaced
The screenshot **06** showed some tools no longer in the image, didn't show some new ones
- with the replacement of  this screenshot, some tools which were no longer there needed removal
- new tools needed a description
- while at it, I sorted the descriptions to fit the order of the icons in the screenshot

Also **added a warning** to expect an ECDSA fingerprint error message when connecting to a rescue'd server due to temporary system with equally temporary ssh server


OVH-Team-Guides:
Merge done, thanks for contributing !

PR #194, opened 2019-04-15
Update guide.en-gb.md: wrong query URL, corrected

NeodymiumFerBore:
The example query uses https://eu.api.ovh.com/1.0/domains/, which fails with the error {"message":"Got an invalid (or empty) URL"}
This does not exist, the correct query is : https://eu.api.ovh.com/1.0/domain


OVH-Team-Guides:
Merge done, thank you for your contribution !

PR #212, opened 2019-05-10
CAT3 updates 

talmortis:
**bridge_client**
added screenshots for vmac configuration in proxmox and vmware

many small corrections. recreated configuration in most of the adjusted OS

was unable to reproduce Arch-Linux guide. but the existing one was wrong on too many levels
(connect to a vm which is currently offline, install something from a mirror, to do so use apt.... (packet manager in arch is pacman) use this tool...)
in a potential future edition, a new part for arch linux could be added, ideally also tested

**network_ipv6**
replaced "use ssh....." with "connect to your server + guide link", becaue basic ssh and shell knowledge is listed as requirement for the guide. 
changed "reboot server" to "restart network"
removed false requirement: ipv4 with fail over (only required if used in vms, see bridged)

**raid_hard**
added check commands to verify the status of the raid controller and its battery.


OVH-Team-Guides:
Merge done, thank you for contributig !

PR #259, opened 2019-06-12
Add a new guide to boot instance with pub and priv

arnaudmorin:
Add a new guide to boot an instance with both public and private
network.

Signed-off-by: Arnaud Morin <arnaud.morin@gmail.com>


arnaudmorin:
WIP
I need to find / create both:
- order vrack
- associate vrack to project

PR #299, opened 2019-07-18
Move new Analytics Data Platform guides to Public Cloud

Kureeru:
As we prodded Analytics Data Platform, we updated the user documentation and moved it from labs to Public Cloud universe.


OVH-Team-Guides:
Files are incorrectly formatted. Exemple : no slug; meaning no URL generation during the build.

PR #304, opened 2019-07-26
Add managed databases EOL policy and announcements

mdavranche:


baaastijn (approved):
text reviewed, good for me.

PR #392, opened 2019-11-14
Precision about region name case in the aws config

jbarotin:
aws cli not work with a region in lower case


OVH-Team-Guides:

PR #394, opened 2019-11-14
Update starting_with_swift_s3api

edausq:
adding precision on "region" configuration, that should be uppercase


edausq:
cloned by #392

PR #402, opened 2019-11-25
Misleading grave accent on dashboard proxy URL and versions incompatibilities.

iakko:
There is a misleading grave accent (`) at the end of the dashboard local URL that results in a missing page while requestes through copy-paste.


iakko:
I've also changed documentation related to K8S Dashboard version to use according to cluster version.

LostInBrittany (approved):

OVH-Team-Guides:
Merge is done, thanks for your contribution

PR #455, opened 2020-02-12
emphasis on the config requirement

ticapix:
emphasis on the configuration requirement upper vs lower case


OVH-Team-Guides:

PR #461, opened 2020-02-14
GB/FR(Logs Data Platform): add information on size about Index As A Service

jehuty0shift:
This commit add a paragraph about Index Size  monitoring of Index As a Service on Logs Data Platform 

Signed-off-by: Babacar Diassé <babacar.diasse@corp.ovh.com>


OVH-Team-Guides:

PR #470, opened 2020-03-03
Fix region name in aws config file

jlutran:
- Fixed an missing quote in curl POST payload
- Public Cloud region name can now be set in lower or upper case.

Signed-off-by: Julien Lutran <julien.lutran@corp.ovh.com>


OVH-Team-Guides:

PR #480, opened 2020-03-13
Update s3ql documentations

edausq:
Current documentation is using keystone version 2 instead of keystone version 3. But keystone version 2 is going to be shut down in 10 days.


OVH-Team-Guides:

PR #486, opened 2020-03-19
Fix typo with command table

obedouet:
Command table which list available options was not properly formatted


OVH-Team-Guides:

PR #490, opened 2020-03-23
FIX url in starting_with_swift_s3api

hparfr:
$OS_AUTH_URL already contains "v3", so users get {"error": {"message": "The resource could not be found.", "code": 404, "title": "Not Found"}}


OVH-Team-Guides:

PR #524, opened 2020-04-20
FR adding hidden tag onto deprecated phones guides

Y0Coss:
Please do not merge before prodding of new phones range 
Has to be prodded simultaneously with New guide : Nos précédentes gammes de téléphones


OVH-Team-Guides:
Internal merge

PR #528, opened 2020-04-23
Delete data-compute lab, it's now in beta as data processing

BellionBastien:
Data-compute guides are now outdated as we launch the beta of data processing


OVH-Team-Guides:

PR #529, opened 2020-04-23
WWW2020 - 1st batch - Hosting, DNS & Email

OVH-Team-Guides:
To proofread (links edition)
Only exported to en-ca, fr-ca, us-es
Not to be prodded before product launch


OVH-Team-Guides:
Internal merge

PR #546, opened 2020-05-07
Update for the Log Data Platform doc (syslog-ng conf)

Frzk:
Modified the template in the config example in order to make it fit RFC5424.

RFC5424 states that the first field must be the **Priority** (between an opening `<` and a closing `>`). The Priority is defined as a combination of both the **Facility** and **Severity** ("The Priority value is calculated by first multiplying the Facility number by 8 and then adding the numerical value of the Severity.")

In syslog-ng, this corresponds to the `${PRI}` macro.

The (previous) `${LEVEL_NUM}` macro only represents the **Severity** (in RFC5424 terms). The syslog-ng doc is really misleading here because it states : "The priority (also called severity) of the message".

I also updated a bit the comments above the template so it's easier to understand for people outside of OVH.
And updated some not-so-important syntax things so that it corresponds to what we can read in the official doc.


OVH-Team-Guides:
Merge is done on develop, thanks for your contribution !

PR #555, opened 2020-05-13
K8s - Velero guide - i18n

LostInBrittany:


OVH-Team-Guides:

yann2192:
Hello, current doc is not working.

The correct url for velero should be `https://s3.<public cloud region without digit>.cloud.ovh.net` and not `https://storage.<public cloud region without digit>.cloud.ovh.net`. 

There is another issue : https://gitter.im/ovh/kubernetes?at=5ee272721e099b0388be57ac

I can make a PR if needed to fix that.

LostInBrittany:
The Velero doc is under refactoring. A new version will be released by end June, alongside with a simpler solution for snapshots, using Stash.

We needed to wait for K8s 1.18 CNCF certification. I am working in those two docs now

PR #557, opened 2020-05-14
fix(Metrics): fix protocols and influx time precision docs

aurrelhebert:


OVH-Team-Guides:

PR #575, opened 2020-05-28
Enhance instructions

adamo-ovh:
Update image to English and also add extra steps for assistance such as how to set the root password after first login. Also explain if the URL expired the customer needs to reinstall now.


OVH-Team-Guides:
Merge is done on develop, merge on master will be done today. Thanks for your contribution.

PR #587, opened 2020-06-08
www2020 universes/products pages

OVH-Team-Guides:
Not to merge before the launch of Web products in America / APAC


OVH-Team-Guides:
Internal merge

PR #620, opened 2020-07-08
[hosting] Adding cluster051

rclsilver:
we just opened cluster051 at BHS


OVH-Team-Guides:
Thanks for your contribution, merge is done on develop. Merge on master will be done shortly.

PR #627, opened 2020-07-13
Private Cloud 2020 updated guides (EN)

bs-ovh:
Adding GB-EN translations of FR updated guides


bs-ovh:
EN update for Private Cloud

PR #686, opened 2020-08-11
EN FR Add MariaDB 10.3 | .4 | .5 EOL

carmelo42:
MariaDB 10.3 | .4 | .5 are availables to our customers. As the EOL dates are not yet known, they are indicated as "To be defined".


OVH-Team-Guides:
Thanks for your contribution @carmelo42, merge is done on develop, merge on master will happen soon.

PR #694, opened 2020-08-18
EN FR Add Mysql 8.0 EOL

wiseflat:
Mysql 8.0 is available to our customers. As the EOL date is not yet known, I set it to  "To be defined".


carmelo42 (approved):
Thanks for your contribution @wiseflat , merge is done on develop, merge on master will happen soon.

PR #736, opened 2020-09-04
Update VSAN guide.fr-fr.md

fesz69:
Change because of Premier launch.
Has to be done for each translated guide


OVH-Team-Guides:
Thanks for your contribution @fesz69 . Merge is done on develop, merge on master will be done soon.

PR #756, opened 2020-09-11
fix typo in subinterface declaration eth0:0

Romain7495:
Signed-off-by: Romain Labat <romain.labat@smood.ch>
Hello,

This PR is aim to fix a typo in the French version of the doc.
Instead of using `eth0.0` concerning IP aliasing for dedicated server, we need to use `eth0:0` to declare an interface
This is well documented here : https://wiki.debian.org/fr/NetworkConfiguration

Kind regards


OVH-Team-Guides:
Thanks for your contribution @Romain7495. This fix is now applied on develop, it will be published online today.

PR #788, opened 2020-09-23
Qemu agent instructions

gary-ovh:
As some users reported that a reboot is required with some distros. Added step to have all users reboot and then verify that the service started.


bs-ovh:
Thanks, Gary! :)

PR #794, opened 2020-09-28
Update list of impacted POP by Arista bug

obedouet:
London and Ashburn were missing from POP list impacted by Arista bug.


OVH-Team-Guides:
Thanks @obedouet, merge is done on develop, we will merge it on master soon.

PR #844, opened 2020-10-19
Get egress IPs from annotation

sguyennet:


bs-ovh:
Thanks for your contribution, sguyennet!

PR #846, opened 2020-10-21
Added info about CephFS

sadziu82:


sadziu82:
Found some typos, will create another PR.

PR #896, opened 2020-11-07
Fix listing images command as it should use 'glance'.

GhaziTriki:


OVH-Team-Guides:
Thanks for your contribution @GhaziTriki. Merge is now done on develop, final merge on master branch will happen soon.

PR #921, opened 2020-11-18
Cloud: AI Training: Add documentation

co42:
This is the document for the AI Training product which should be in production soon.


OVH-Team-Guides:
Thanks @co42 , as agreed me move your PR to stash.

PR #990, opened 2020-12-10
Fix debian example

benharri:
Debian uses iproute2 now and should be updated in the guest config example


OVH-Team-Guides:
Hi @benharri 
Thanks for this contribution. As we must also consider older distributions which may not use iproute2 yet, we integrated your updated version into a new complete one, mentioning older and latest distributions.
Please see https://github.com/ovh/docs/pull/996
Best regards, 
The OVHcloud guides team

PR #1003, opened 2020-12-14
[UpdatePhp] Reduce of php timedout, task 47717

TheGoat-ma:
We have perform an update on PHP configuration to reduce php timedout from 300s to 165s. So we need to update our guides


OVH-Team-Guides:
Thanks for your first contribution @TheGoat-ma . Merge on master will be done very soon.

PR #1170, opened 2021-02-23
Update slug FR to EN

YoannAtOVH:
link in EN


OVH-Team-Guides:
Thanks for your contribution @YoannAtOVH . Merge is done

PR #1266, opened 2021-03-23
Hiding guide

gary-ovh:
Option not available in subs managed by CA


OVH-Team-Guides:
Thanks for your contribution @gary-ovh. Sorry that this wasn't checked before. Merge on master will be done today.

PR #1355, opened 2021-04-09
Update guide.fr-fr.md

haeiven:
Retirer le 0 et le 1 après avoir entré l'adresse IP pour debian, cela génère une erreur.

Error: either "local" is duplicate, or "0/255.255.255.255" is a garbage.


bs-ovh:
Hello @haeiven,
We are currently updating guides related to IP configurations in all languages. Since your correction is included in a larger update, we are closing this request. Thank you for contributing!

PR #1363, opened 2021-04-13
Fix image link on ZyxelxOTB WiFi

AxelDum:


OVH-Team-Guides:
Thanks for the fix @AxelDum !

PR #1404, opened 2021-04-23
PUPI-540: Changes in troubleshoot guide

somi99-byte:
Following changes are addressed:

- in the MySQL server has gone away section, the disk space issues are missing.
- in the “Read-only file system” error section, the sentence "During the build phase of your application, the main filesystem is writable. So you can do whatever you want (e.g. compile code or generate anything you need). But during and after the deploy phase, the main filesystem will be read-only." is missing. (even if it's without the links)


OVH-Team-Guides:
Thanks for your contribution @somi99-byte, merge on master will be done soon.

PR #1470, opened 2021-05-07
FR update Monitoring For Telecom XDSL

timig:
Update guide Monitoring For Telecom XDSL.
Update with screen and last manager use.


OVH-Team-Guides:
Thanks for your contribution @timig . Merge is done on develop. Merge on master will be done very soon.

PR #1567, opened 2021-06-04
PUPI:558 Changes in configuration-services guides

somi99-byte:
Including endpoint description from PSH guides


OVH-Team-Guides:
Thanks for your contribution @somi99-byte, merge on master will be done soon.

PR #1669, opened 2021-06-29
Update for the launch of D2

mhurtrel:


OVH-Team-Guides:
Thanks for your contribution @mhurtrel, duplication will be done and deployment later today.

PR #1756, opened 2021-07-23
Fix table to be 2 column instead of 3

AxelDum:


OVH-Team-Guides:
Thanks for your fix @AxelDum, we'll deploy it soon.

PR #1769, opened 2021-07-27
prefer https by default

petrus-v:
As far my understanding there is no HTTP -> HTTPS redirection using documented url:
`http://www.ovh.com/nic/update?system=dyndns&hostname=$HOSTNAME&myip=$IP`

So I suppose that user/password travel in clear text on the network, is that a security issue
(by Man in the middle attacker)?

I supposed that a lot of former application do not support redirection that why this support still present
but I would recommend to make it deprecated anyway and at least use HTTPS by default in this documentation.


OVH-Team-Guides (approved):
Hello @petrus-v, sorry about that unusual delay, we had to check if the https link was usable.
Thanks for your contribution.

OVH-Team-Guides:
Thanks for your contribution @petrus-v. We will deploy it soon.

PR #1816, opened 2021-08-17
FR updating Fax configuration in Control Panel for Telecom Fax

Satchurated-RBX:


OVH-Team-Guides:
Thanks for your contribution @Satchurated-RBX, deployment will happen soon.

PR #1817, opened 2021-08-17
FR updating Sending SMS via Control Panel for Telecom SMS

Satchurated-RBX:


OVH-Team-Guides:
Thanks for your contribution @Satchurated-RBX, deployment will happen soon.

PR #1853, opened 2021-08-26
FR updating Contacts Book for Telecom VoIP

Milytum:


OVH-Team-Guides:
Thanks for your contribution @Milytum, deployment will happen soon

PR #1988, opened 2021-09-28
Hiding guides for unsupported features in CA/APAC

garyyamamoto:
Guides to be hidden. Options not available in Canada


OVH-Team-Guides:
Thanks for your contribution @garyyamamoto, deployment will be done soon.

PR #1993, opened 2021-09-28
Update ISSP FR

GdMnt:
Update french ISSP to align it with EN version.


OVH-Team-Guides:
Thanks for your contribution @GdMnt. Deployment will happen soon.

PR #1994, opened 2021-09-29
Fix/notebook ports

christophe-rannou:
Getting started for notebooks
Debug job for training
Small fixes


OVH-Team-Guides:
There are multiple fixes to do on those pages before PR should be done again.

PR #2057, opened 2021-10-13
fix: error 404 ovh svg

scraly:


OVH-Team-Guides:
Thanks for your contribution @scraly.

PR #2081, opened 2021-10-25
EN/FR [Logs Data Platform] add IP handling information in field namin…

jehuty0shift:
…g conventions

IP adresses are not allowed to be used as key fields in Logs Data PLatform. The commit add this information in the field Naming Convention documentation.


OVH-Team-Guides:
Thanks for your contribution @jehuty0shift, deployment will be done soon.

PR #2177, opened 2021-11-22
Notes on subnet size

obedouet:
Add notes about non-restriction in size on statically routed subnets.


OVH-Team-Guides:
Thanks for your contribution @obedouet.

PR #2277, opened 2021-12-10
Updating EN/FR slugs for S3 Object Storage guides

fpillotovh:


OVH-Team-Guides:
Thanks for your contribution @fpillotovh

PR #2287, opened 2021-12-13
feat: remove unsupported languages

wdhif:
Remove guides in unsupported languages.


wdhif:
Seen with @BrenKt they will manage translation.

PR #2350, opened 2021-12-22
add "Taint, cordon and drain specific Nodes and Nodes Pools" tutorial

scraly:


mhurtrel (commented):
We previously show => We previously showed
Nos is unreachable => Node is unreachable
unaivalable => unavailable
about a particular Node Pool. = within a particular Node Pool.
Cordon a node means make the node unschedulable. =>
Cordoning a node means making the node unschedulable.


Instead of cordon each node manually: => Instead of cordoning each node manually:


will respect the `PodDisruptionBudgets` you have specified (if relevant). => I would add a sentence or 2 about this and/or link to the kubernetes.io doc about this ?

I propose that change (but feel free to challenge me if wrong)  :
Now you can do whatever you want on Nodes or on this node pool and after that you can uncordon the Nodes of the node pool, to come back as before the drain
=>
As shown in the console output, draining a node also cordons its. You can do whatever you want on Nodes as you are sure they do not run any workloads. Do not forget to uncordon the Nodes of the node pool, to allow workloads to be run again:
else they can’t be scheduling => additionally to not being schedulable. 

For clarification on the last use cas, i would make the 2 nodepools with different flavors (for example d2-8 and b2-7)

scraly:
Fixed @mhurtrel remarks :)

OVH-Team-Guides:
Thanks for your contribution @scraly and @mhurtrel

PR #2388, opened 2021-12-30
Remove mentions of old LBaaS FQDNs in k8s docs

XavierDuthil:
Some docs are outdated with information about the old LBaaS mechanism (before the switch to IOLB).


OVH-Team-Guides:
Thanks for your contribution @XavierDuthil

PR #2420, opened 2022-01-07
Modify bond-xmit_hash_policy see PUBM-17875

dbebel:


OVH-Team-Guides:
Thanks for your contribution @dbebel

PR #2456, opened 2022-01-14
Fix misspelling in console display

scraly:


OVH-Team-Guides:

PR #2460, opened 2022-01-17
aws config profile name for default changed

x0ptr:
according to the official doc it should be [default] not [profile default].
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

test with:
aws-cli/1.22.37 Python/3.9.7 Darwin/21.1.0 botocore/1.23.37

[profile name] is possible but not [profile default]
aws cli will not detect the config.


XavierDuthil (approved):
Approved, thank you for your contribution

OVH-Team-Guides:
Thanks for your contribution @larsonnn

PR #2497, opened 2022-01-28
Updating FR OVHcloud Load Balancer guides

gayeamdy:


OVH-Team-Guides:
Common acknowledgment with you @gayeamdy that this PR is too anticipated. Closing if for the moment.

PR #2560, opened 2022-02-09
Fixing location of API call in vDC migration for Hosted Private Cloud

Hubmar:
Corrected the wrong section SK 706


OVH-Team-Guides:

PR #2587, opened 2022-02-15
Fix PV example

MaximeWeyl:
The mount path is not correct, as documented here : 
https://hub.docker.com/_/nginx

This makes the example not runnable (or at least no as expected)


OVH-Team-Guides:
Thanks for your contribution @MaximeWeyl , we will duplicate it on other versions.

PR #2627, opened 2022-02-23
PQ - Obsolete Removal and vROPS intro - En & Fr

PierreHenriWrites:
VM freeze update En & Fr

Deletion off all vROPS existing docs as obsolete and without value as is

Creation of the vROPS intro to replace all that was deleted. Everything deleted should be redirected to the new one.


OVH-Team-Guides:
Thanks for your contribution @PierreHenriWrites. Redirections have been written

PR #2660, opened 2022-03-02
:memo: add bash coloring syntaxing

philippart-s:
To fix the `_helper.tpl` coloration in the tutorial


OVH-Team-Guides:
Thanks for your contribution @philippart-s

PR #2689, opened 2022-03-08
add update ACL info

raclepoulpe:
Because updating ACLs is only allowed via API, it's important to specify this.


OVH-Team-Guides:
Thanks for your contribution @raclepoulpe

PR #2701, opened 2022-03-09
feat(clouddb): add sql_mode configuration

Meuh-42:


Meuh-42 (commented):
There should not be spaces between list of options.
It will not be displayed with spaces in the manager, as spaces will break the configuration

bs-ovh (approved):

PR #2837, opened 2022-04-06
fix: add missing sudo in migration commands

franckduriez:
You need to be root to modify the files in /etc/apt
Without sudo, the commands are failing


bs-ovh (approved):

OVH-Team-Guides:
Thanks for your contribution @franckduriez

PR #2913, opened 2022-04-25
Updating default filename on Kubectl config for K8s

mhurtrel:
Update the command line to config KUBECONFIG env variable to match default  file name when downloading from the UI/API


OVH-Team-Guides:
Thanks for your contribution @mhurtrel

PR #3015, opened 2022-05-16
Updating Network Virtual Mac assignment for Dedicated Servers

dbebel:
Add call API for add / remove a virtual MAC on IP


OVH-Team-Guides:
Thanks for your contribution @dbebel

PR #3059, opened 2022-05-30
Split exposed api doc to two differents: API and plugins versions

scraly:


OVH-Team-Guides:

PR #3199, opened 2022-06-27
Create tuto how to secure a nginx ingress

scraly:
change orders
and route cert-manager tuto to this new guide instead of old not relevant one


OVH-Team-Guides:

PR #3256, opened 2022-07-18
use MongoDB unified URL format throughout the docs

rgbj:
MongoDB services now support connections through a single unified URL, so reflect that in the docs


OVH-Team-Guides:
As agreed with @fpillotovh , this content will be integrated into another PR. Thanks for your contribution @rgbj

PR #3339, opened 2022-08-02
Fixed gateway ip

ImMarvolo:
In the documents it says that the gateway ip is the one ending in .254 but that is not true as it would not route your traffic. The correct gateway ip is the one ending in .252.


Jessica41:
Hello @ImMarvolo , thank you for your contribution. After verifying on our end, our configuration remains the same and works with the gateway ending with .254. If you have faced or are facing routing issues using this configuration, may I suggest creating a support ticket so our support teams can look into it?

PR #3348, opened 2022-08-04
Fix k8s managing nodes with api

XavierDuthil:
The documentation on how to manage nodes via API was clearly outdated.

Also applied a small update to the known limits about the number of volumes that can be attached to an instance.


OVH-Team-Guides:
Thanks for your contribution @XavierDuthil

PR #3361, opened 2022-08-11
No longer limited

mhurtrel:
Live snapshots can now be performed on all regions


Jessica41:
Thanks for your contribution @mhurtrel!

PR #3370, opened 2022-08-19
duplicate EN content on FR for ML serving EOL notice

Jessica41:


OVH-Team-Guides:
@Jessica41 , the ML Serving documentation only exists in English, so retiring this PR.

PR #3441, opened 2022-09-06
update FR et DE

Kazarus4:
Updates :

- Anchor links gofurther=>go-further
- Anchor link Aller-plus-loin =>go-further to the uniformization
- Replace the updates which was disapear in DE version


Kazarus4:
See with Yoann to merge

PR #3506, opened 2022-09-29
fixing the SPF with the correct server

barbieri1982:
For Canada service the correct SPF is mx.ovh.ca and not mx.ovh.com


OVH-Team-Guides (approved):

OVH-Team-Guides:
Thank you for your contribution @barbieri1982. We'll duplicate these fixes onto the other concerned versions.

PR #3558, opened 2022-10-12
fix(LDP): update tokens related informations - EN and FR

jehuty0shift:
This PR update Logs Data Platform documentation about tokens. We add the possibility to use tokens in a more conventional way in our platform (password being the token and not the opposite), so we changed our examples  to reflect that.


OVH-Team-Guides:
Hello @jehuty0shift , unless you have a specific reason for not updating it, I suggest you update the "Last updated" date at the top of the guide.

jehuty0shift:
Hi, thank you. You're right, i pushed force a commit to also change it.

OVH-Team-Guides:

PR #3563, opened 2022-10-13
fix(IPLB): 307 instead of 303

Alkorin:
Fix typo


OVH-Team-Guides:
Thank you for your contribution @Alkorin

PR #3573, opened 2022-10-14
Correct mail server

barbieri1982:
ssl0.ovh.net is not good for CA email service.


BrenKt:
Regroup in https://github.com/ovh/docs/pull/3585

PR #3574, opened 2022-10-14
Correct webmail

barbieri1982:
There is no roundcube for CA email service, only OWA.


OVH-Team-Guides:
Thank you for your contribution @barbieri1982

PR #3575, opened 2022-10-14
Correct MX and SPF server

barbieri1982:
Setting the correct MX record and SPF record


OVH-Team-Guides:

PR #3576, opened 2022-10-14
correct path and label

barbieri1982:
On CA manager exchange is under "Bare Metal Cloud" section and the label in the tree to click is "Microsoft Exchange".


OVH-Team-Guides:
As mentioned to you @barbieri1982, this path issue will be fixed by the concerned team.

PR #3577, opened 2022-10-14
Correct webmail link

barbieri1982:
on CA the correct webmail is https://ex.mail.ovh.ca/ and not ex.mail.ovh.net


OVH-Team-Guides:

PR #3580, opened 2022-10-14
Change .uk link with CA link

barbieri1982:
Fixing some URL targeting .uk domain instead of https://www.ovhcloud.com/en-ca/


OVH-Team-Guides:

PR #3581, opened 2022-10-14
EN-GB: add section about service account tokens to installing-kubernetes-dashboard

MylesDev:
From versions of Kubernetes v1.24.0 onwards, a new feature `LegacyServiceAccountTokenNoAutoGeneration` is in beta, and enabled by default. This feature, when enabled means Secret API objects containing service account tokens are no longer auto-generated for every `ServiceAccount`.

Because of this, the current guide for getting a bearer token will no longer work. The documentation added creates a token for the service account (as it is now needed), so that a bearer token can be retrieved for authentication


LoicBlo (commented):
works for me too so fine to merge it :)

OVH-Team-Guides (approved):

OVH-Team-Guides:
Thank you for your contribution @MylesDev, we will soon do some minor updates on your work.

PR #3568, opened 2022-10-13
Updated PCI/API Nova (Nova CLI Depreciated)

al26p:
Nova CLI is deprecated and will be a removed in a future release.

Updating doc to use Python OpenStack CLI.

PR #3591, opened 2022-10-17
Updating doc to use Python OpenStack CLI.

al26p:
Nova CLI is deprecated and will be a removed in a future release.

Updating doc to use Python OpenStack CLI.

(Follow up of #3568 )


OVH-Team-Guides:
Thank you for your contribution @al26p 
Translations will be done soon

PR #3652, opened 2022-11-03
Fix panel display in known limit doc

LoicBlo:
I've seen some display issue in the known limit doc since my last PR.
This push will fixed it.


OVH-Team-Guides:
Thank you for your contribution @LoicBlo

PR #3680, opened 2022-11-10
fix(isp): new password is send by email

AxelDum:
Inform customer that in case of a transition from managedByOVH to non managed, a new password is generated and send by email


OVH-Team-Guides:

PR #3685, opened 2022-11-14
fix(overthebox): add link to older usb way

AxelDum:


OVH-Team-Guides:

PR #3695, opened 2022-11-15
fix(LDP): add a new configuration for new Container Runtimes

jehuty0shift:
Hi, 
Following issues with customer regarding Kubernetes with fluent bit configuration, we are updating the fluent bit documentation in order to reflect the new Container Runtime options in Kubernetes


OVH-Team-Guides:

PR #3705, opened 2022-11-16
ES - Fix link to support offer

GSIM-VOC:
syntax error in line 401


OVH-Team-Guides:
Thank you for your contribution @GSIM-VOC

PR #3735, opened 2022-11-18
Web Cloud guides updates

barbieri1982:
Guided that need to be deleted in CA
added info for external domain on hosting guide


OVH-Team-Guides:
Thank you for contribution @barbieri1982

PR #3768, opened 2022-11-28
Managed Private Registry is HDS

mhurtrel:
Since Spring, the service in in HDS and ISO perimeters Please spread accross all languages


OVH-Team-Guides:
Thank you for your contribution @mhurtrel

PR #3791, opened 2022-12-01
Updating the "Last updated" date after a full review

luigi-p:


OVH-Team-Guides:
Hello @luigi-p , thanks for the review. As agreed with you, let's work this another way.

PR #3806, opened 2022-12-05
Update NAS NFS guide

gary-ovh:
Added "/" to path as required to avoid confusion with customer.


OVH-Team-Guides:
Thank you for your contribution @gary-ovh

PR #3915, opened 2022-12-20
Additional IP : add section + add cancellation guides (EN,FR)

thieummm:
Hey Yoann,

please consider this.
I just have a problem with the order of the section, it should be at the top. Could you please fix it so it is displayed at top left of the page ?


OVH-Team-Guides:
Thank you for your contribution @thieummm

PR #3925, opened 2022-12-22
KS--SK1000--Full_Update-geolocation_web_hosting-Translation

Kazarus4:
Translations are done for all subs / links checked with gitpod and all are in "200 OK" status.

Need peer review for EN, latin and PL version (DE version will be completly check in a second time).


Kazarus4:
https://8080-ovhclouddev-docsrenderi-w0n7suvezna.ws-eu80.gitpod.io/ => rendering check and all is ok.

Just need Proof now EN, latin and PL version (DE version will be completly check in a second time).

PR #4007, opened 2023-01-11
Fix hello world link in the tuto 

scraly:


OVH-Team-Guides:

PR #4008, opened 2023-01-11
Update velero tuto with new versions and add more commands

scraly:


OVH-Team-Guides:

PR #4139, opened 2023-02-01
All languages update - Speech to text app for AI Deploy

MathieuBsqt:
Here is a small update to remove a packages.txt file from the documentation that is not required anymore by the speech to text app.


OVH-Team-Guides:

PR #4140, opened 2023-02-01
HM-2DS mini

Hubmar:
adding a minimum of 2 DS per PCC


OVH-Team-Guides:
Thank you for your contribution @Hubmar

PR #4160, opened 2023-02-08
Fix typo in kube proxy doc

scraly:


OVH-Team-Guides:
Thanks for the fix @scraly

PR #4242, opened 2023-02-17
[ai notebooks] - tuto speech to text - links update

MathieuBsqt:
Conda framework has been renamed Miniconda. This change has had an impact on some links.

Here is an update so that they are still clickable.


OVH-Team-Guides:

PR #4253, opened 2023-02-21
Adding link for the HG and Scale configuration help

belug23:
The HG and Scale servers aren't on the same network as the others and this documentation doesn't work. Adding the information about those servers exception and the link to the Additionnal IP configuration page for the HG and Scale server will probably help direct people to the right page.


bs-ovh:
Thank you @belug23 !

bs-ovh (approved):

PR #4257, opened 2023-02-23
KS-SK1015-TRANSLATIONS-TUTO-WOOCOMERCE-WP

Kazarus4:
All translations are done.

links with gitpod => 200 ok
render with gitpod => all seems to be good

Need proof for DE, EN, latin and PL version


OVH-Team-Guides:
ES-IT-PL-PT OK for prod

PR #4291, opened 2023-03-01
Update Keystone token manipulation

edausq:
Hello,

http headers should be treated as insensitive.

Following Keystone upgrade, header has been update from from X-Subject-Token to x-subject-token.

This patch add option to ignore case in the example scripts/commands.

Edouard


OVH-Team-Guides (approved):

OVH-Team-Guides:
Thank you for your contribution @edausq

PR #4294, opened 2023-03-02
Fix an uncropped image of the Compass window

thild42:
Fix an uncropped image on the MongoDB - Understand Connection Strings guide.


OVH-Team-Guides:

PR #4304, opened 2023-03-03
Add a warning regarding mandatory ACL to load sample data

thild42:
Following a question on Discord from an Opensearch user, we add a warning to the Getting Started Opensearch guide telling users that specific ACL needs to be defined in order to use the sample data.


OVH-Team-Guides:

PR #4343, opened 2023-03-10
feat(ldp): Cold storage encryption

pfrayer:
Hello :wave: 

In OVHcloud Logs Data Platform, there's an existing feature named "Cold storage" ([cf. this doc](https://docs.ovh.com/fr/logs-data-platform/cold-storage/)).

On top of this, we added a new feature named "Cold Storage Encryption", which allow users to encrypt their cold storage archives. This PR contains the documentation about this new feature. 

We (LDP team) preferred to add a dedicated documentation page for this new feature instead of improving the existing Cold Storage documentation page, to avoid having an over-weighted doc page.

I also re-ordered the pages in the "[Features](https://docs.ovh.com/fr/logs-data-platform/)" section so both the existing doc and the new one are one-after-the-other in this Features section, that's why the `order` of others pages is modified.


OVH-Team-Guides:
Thank you for your contribution @pfrayer

PR #4366, opened 2023-03-15
doc: 📝 add some precision on parameters

philippart-s:
Add some precision about how to set custom parameters in the chart and especially the possibility to not install some components.

@scraly tell me if it's not clear enough, we can cancel this PR if it's.


pages/platform/kubernetes-k8s/monitoring-apps-grafana-prometheus/guide.en-gb.md
scraly, line 1: don't forget to update the "last updated" date and the "updated" meta :)

scraly (changes requested):

PR #4370, opened 2023-03-16
Hb raci object storage

H4516022Y:
docs / pages/ platform / public-cloud / responsibility model object storage

hello,

RACI in vFR and EN to be published for Object storage product , only for specific range : Standard Object Storage-S3 API, High Performance Object Storage- S3 API and Standard Object Storage-Swift API .

can please add links in the following page too ? : https://docs.ovh.com/fr/partage-responsabilite/

thanks for your actions


OVH-Team-Guides:

PR #4375, opened 2023-03-16
doc: 📝 add some precision on parameters

philippart-s:
Add some precision about how to set custom parameters in the chart and especially the possibility to not install some components.

branch was renamed so this is the same PR that #4366


philippart-s:
@scraly for me it's done 😉

scraly (approved):
Perfect! :)

bs-ovh (approved):

PR #4400, opened 2023-03-23
fix: split hosts and zpools/datastores

YoannAtOVH:
Clarification between Host and Datastore


OVH-Team-Guides:

PR #4456, opened 2023-03-31
Public Cloud Databases: update image names of the restore backups guide

thild42:
As a workaround to force refresh of images, image names have been updated.


OVH-Team-Guides:

PR #4492, opened 2023-04-11
Dev/mferon/clouddb add new dbms

mferon:
Hello,

We are adding MariaDB 10.11 and PostgreSQL 15 to CloudDB, so we also add them in the EOS/EOL policies :)


Jessica41:
Thank you for your contribution @mferon

PR #4501, opened 2023-04-13
add deprecation warning on odp metrics

Kureeru:
Added a deprecation warning on the features related to the get metrics endpoint of data processing as it is going to be removed from API


bs-ovh:
Thank you @Kureeru !

PR #4543, opened 2023-04-26
Update deploying gpu app doc to be compliant with ubuntu 22 nodes migration

scraly:


OVH-Team-Guides:

PR #4570, opened 2023-05-02
clarify outbound public IP in Gateway FAQ after feedback from customer

yomovh:


OVH-Team-Guides:

PR #4573, opened 2023-05-03
Delete FAQ (no longer needed)

remrozk:


OVH-Team-Guides:
We need to write related 301 rules to avoid 404s for our customers.

OVH-Team-Guides:
Thank you for your contribution @remrozk. Redirections are written

PR #4610, opened 2023-05-10
Dev/alimoge/intervention without blur

ArnaudFromGitHub:
Add blur inside gifs files.


OVH-Team-Guides:
Replaced with https://github.com/ovh/docs/pull/4611

PR #4619, opened 2023-05-11
Update bond configuration LB and apply proxy_arp only on public bond only

ccleouf66:
Update bond configuration LB and apply proxy_arp only on public bond only


OVH-Team-Guides:
Thank you for your contribution @ccleouf66 
We will update translations accordingly

PR #4698, opened 2023-05-26
KS-SK1129-WHGR-Update-htaccess_protect_directory_by_password-for-Webhosting

Kazarus4:
Full update and new scripts based on Apache recommandations and latests apache versions 2.4.

FRFR done, need tech rv before peer rv.


Kazarus4:
tech rv done, need peer rv before merge FR

PR #4699, opened 2023-05-26
Object Storage - S3 website HTTPS - Minor fix for FR versions

lason-ovh:
undo translation of Host to Hôte in FR guide (Host should not be translated because it is the name of the standard http header)


OVH-Team-Guides:

PR #4724, opened 2023-06-01
Object Storage  S3 - Limitations - Update

lason-ovh:
Clarifications about maximum object sizes that can be uploaded


OVH-Team-Guides:

PR #4729, opened 2023-06-01
fix (SSO with google workspace) - Fix a screenshot with bad informations

VincentCasse:
The two fields was reversed.


OVH-Team-Guides:
Hello Vincent, even tough it is only a screenshot update, I suggest we also update the `updated:` key in the guides header, so that our customers know that content has been updated today.

VincentCasse:
Of course, I forgot this point. PR updated

OVH-Team-Guides (approved):

OVH-Team-Guides:
Thank you for your contribution @VincentCasse

PR #4734, opened 2023-06-05
Fix bad formatting in the Kafka MirrorMaker capabilities guide

thild42:
"Nodes" title was merged with the info panel just before it because of a missing line feed.


OVH-Team-Guides:
Thanks for your contribution @thild42

PR #4779, opened 2023-06-12
(fix) host api call

ThibautOVH:
Fix host API call (host, not filer)


OVH-Team-Guides:
Thank you for your contribution @ThibautOVH

PR #4805, opened 2023-06-16
KS-SK1157-KFT0004337-Update-Change-domain-owner

Kazarus4:
Updates =>

go further,

Content of 2 or 3 paragraphs and typo

need peer rv for FRFR version.  After, I think they'll need a tech rv to check if all are still ok. This before continue to duplicate for FRCA and translations.


Kazarus4:
cancel PR because content merged and worked on previous PR 4803.

PR #4813, opened 2023-06-19
Update LDP inputs guide: document input autoscaling

pfrayer:
We added a new feature regarding LDP Data gathering tools (often called "inputs"): it can now be auto-scaled.

As a doc page already exists for these inputs, we simply update this page to quickly explain the new feature (and update the screenshot)


OVH-Team-Guides:
Thank you for your contribution @pfrayer

PR #4829, opened 2023-06-21
Add rate limits info for OpenStack API

arnaudmorin:
We will introduce rate limits on most of the OpenStack API. This documation will help user figure our why and how they can avoid hitting the limits.


OVH-Team-Guides:
Thank you for your contribution @arnaudmorin

PR #4831, opened 2023-06-22
Public cloud db pgvector

thild42:
Add pgvector to the list of available extension in the PostgreSQL Extension guide.


thild42:
Sorry, messed up the PR with an unwanted commit.

PR #4832, opened 2023-06-22
Add pgvector to the list of supported PostgreSQL extensions

thild42:
Add pgvector to the list of supported PostgreSQL extensions


OVH-Team-Guides:

PR #4840, opened 2023-06-23
uIAM : precision about wildcarding

gbarideau:
Precision about the possibility of wildcarding on the URN


OVH-Team-Guides:

PR #4852, opened 2023-06-27
Public cloud data integration fix binaries names

thild42:
Fix the CLI binaries names to be aligned with the actual binaries names.


OVH-Team-Guides:
thank you for your contribution @thild42

PR #4874, opened 2023-06-30
changed Kubernetes cross-referenced doc title

Le-Gruty:
Juste a doc title change to make more sense in K8s doc context.


OVH-Team-Guides:
Thank you for your contribution @Le-Gruty

PR #4878, opened 2023-07-03
Remove enterprise plan and add module licensing information

thild42:
We could not offer the enterprise plan for Redis because they own the trademark Redis Enterprise. I added specificities of the license model regarding popular modules.


OVH-Team-Guides:

PR #4904, opened 2023-07-07
Public cloud data integration fix cli install guide

thild42:
The authentication process description was missing a crucial step setting the service name corresponding to the PCI project leading to 404 error on CLI calls. The documentation is now up to date.


OVH-Team-Guides:

PR #4906, opened 2023-07-10
Franck COSTA NSX EOL

franckcosta:
updated guide with screenshot, fix date.
Can you fix the name of the documentation too ? 
It's NSX End of Life ( and not End of Support).

Thanks a lot.

PR #4910, opened 2023-07-11
minor update about vxlan choice

ThibautOVH:


ThibautOVH:
need to prod https://github.com/ovh/docs/pull/4906 before this one, if it's still needed, I'll do another PR to modify the guide.

PR #4973, opened 2023-07-31
fix: align both FR/EN Migration guides. fix links, units, etc.

YoannAtOVH:
Fix and Translate before adding new idem


OVH-Team-Guides:

PR #4997, opened 2023-08-07
AI products - S3 update - Specify to use Swift and not S3 with Control Panel

MathieuBsqt:
I have added some warnings in the AI Notebooks Manage data Documentation to warn users that S3 buckets are currently not available for AI Products and then AI Notebooks if they are using the OVHcloud Control Panel. They can only use Swift containers, or they need to use the `ovhai cli` which is compatible with S3 buckets.

I also take this PR to correct the name of an indication, which was not the same as the one shown on a screenshot illustrating the operation. (2nd commit)


OVH-Team-Guides:

PR #5036, opened 2023-08-17
KevS-SK1204-KFT0005964-Update-Dynhost-for-DomainDNS

Kazarus4:
Updates : 

FRFR done => add success paragraph on step 1 about domain without sub-domain.
Add sentence on the board on step 2 in bold about domain without sub-domain.


Kazarus4:
need peer rv

PR #5054, opened 2023-08-22
BGP between 2 PCC NSX-t inside the same vRack

jabamin-ovh:
Guide creation "BGP between 2 PCC NSX-t inside the same vRack" for private cloud


OVH-Team-Guides:
As discussed with you @jabamin-ovh , content is not ready yet for publication so I'm closing the PR. Thanks

PR #5058, opened 2023-08-22
update AI Tools x S3 documentation

MathieuBsqt:
Some users have reported problems with the S3 documentation. Here is an update to resolve their concerns. 

- I have update the .gb version. Other language versions have been deleted. 
- I have modified a screenshot, that was originally in French instead of English.


OVH-Team-Guides:

PR #5080, opened 2023-08-25
feat: reduce image size and fix image thumbnails tags

scraly:


OVH-Team-Guides:
Thank you for the fixes @scraly

PR #5084, opened 2023-08-28
Resize big image in configure OIDC for private registry tuto

scraly:


OVH-Team-Guides:
Thank you for the fix @scraly

PR #5154, opened 2023-09-08
new doc about opensearch api

Le-Gruty:
One new doc page for Observability/Logs Data Platform about the usage of the OpenSearch API to expose logs to third-party tools


OVH-Team-Guides:
Thank you for your contribution @Le-Gruty

PR #5159, opened 2023-09-09
fix-webpass-static-folder-and-links

fpillotovh:


fpillotovh:
Yoann,
we should probably arrange a meeting with Quentin to fix 2/3 errors concerning the bot that checks the links :p

PR #5213, opened 2023-09-20
[fix] - resize images

eleapttn:
Resize images for the following guide: `deploy_tuto_16_fastapi_onnx`


OVH-Team-Guides:
Thank you for the fix @eleapttn

PR #5237, opened 2023-09-25
clarify setup for instance backup transfer

yomovh:
This PR clarifies the guide to stress the fact that the environment does not need to be an instance (a local environment can be used). This is a feedback that have been given by customers.

For that reason, the terminal snippet mentioning `#root@server:~` have been removed. 

Also the wording `datacenter` have been replaced by `OpenStack region` (validated with @JacquesMrz) 

Please duplicate to other language if necessary.


bs-ovh (approved):

tcpdumpfbacke:
Thank you @yomovh !

PR #5239, opened 2023-09-26
VMware on OVHcloud - vrops - deleting a deprecated function from the guide

RealAdaas:
Delete the line Add Account because it doesn't work, the firewall blocks it


OVH-Team-Guides:
Thank you for your contribution @RealAdaas

PR #5267, opened 2023-10-10
Updating k8s version on upgrade exemple

antonin-a:
Quick refresh on k8s versions so the doc does not look outdated


BrenKt:
Thx @antonin-a for your contribution, i will duplicate this update on each subs on `BrenKt-PR5267-Update-K8S-version` branch

PR #5288, opened 2023-10-16
ALL Update Data integration CLI install for Data Integration

francoisovh:
Shell completion : change redirection (>> instead of >) to avoid overwriting original bashrc


BrenKt:
Thx @francoisovh for your fix

PR #5289, opened 2023-10-17
removing beta warnings

gbarideau:
Regarding the coming GA of OVHcloud IAM, we should removing beta warnings from the documentations


BrenKt:
Thx @gbarideau for this update :)

PR #5320, opened 2023-10-25
KevS-SK1265-Update-DMARC-for-DomainDNS-Translations

Kazarus4:
Updates : 

- For EN subs which already created => date of last update and Warning paragraph added a the begining of the guide.

I'll check next week with @BrenKt for the other subs versions which don't existing for this guide (DE, EN-ASIA, EN-AU, ES-ES, ES-US, IT, PL and PT). This because I don't know why these versions don't exist.


Kazarus4:
need proof by @tcpdumpfbacke only for my new paragraph which added before merge.

PR #5329, opened 2023-10-26
KevS-SK1259-Update-PMA-Screenshots-for-WebCloudDatabases

Kazarus4:
Updates : (FRFR only)

- Date of last update
- correct old PMA screenshots
- Give more explanation to connect efficiently to a database hosted in a Web cloud databases with PMA.

No tech review needed because it was tested successfully on my side. I just need a peer review by @BrenKt before merge and translate the updates for the other subs.

I'll create another branch for the 2 other guides which are concerned by the same SK1259.


BrenKt (approved):
Done through VScode

PR #5341, opened 2023-10-31
KevS-Update-CRON-for-Webhosting

Kazarus4:
Updates : 

Date of last update
Correcting typo
Add a paragraph to explain difference between Dys and Days of the week

FRFR done (tech rv and peer rv too) => I'll duplicate to FRCA before merge it


Kazarus4:
FRCA done

I update to the go further too

PR #5346, opened 2023-11-02
fp SK-1285 - Update s3 compliancy

fpillotovh:


Kazarus4:
see with f.pillot and anthonin

PR #5349, opened 2023-11-03
KevS-SK1264-Update-Create-db-for-Webhosting-TRANSLATIONS

Kazarus4:
Translations doing


Kazarus4:
need proof for DE and EN versions by @tcpdumpfbacke , need EN duplication after that. This all before merge on develop.

bs-ovh (approved):

PR #5373, opened 2023-11-10
KevS-SK1290-KFT0009251-Update-Install-CMS-1click-for-Webhosting

Kazarus4:
Add two paragraph to sensibilize customers about the "root folder".

This after a cust feedback about this.

Need peer rv and maybe a quick tech review before merging and translations


BrenKt (approved):
RAS

PR #5460, opened 2023-12-01
Add required step to configure UPN for Azure AD SSO doc

Stoakes:
This PR follows a few support ticket by customers.

We missed the UPN (User Principal Name) in Azure AD documentation.
Since Microsoft renamed the product to entra ID since then, I updated title, excerpt and a screenshot as well.


Stoakes:
Closing in favour of #5483

PR #5502, opened 2023-12-08
KevS-SK1298-Update-Create-Shared-SQL-for-WebHosting-TRANSLATIONS

Kazarus4:
All translations are done.

Need proof for : 
- DE and EN by @tcpdumpfbacke 
- Latin and PL by @Y0Coss


Kazarus4:
@Y0Coss  approved these changing. It stay only proof for DE and EN.

bs-ovh (approved):

PR #5506, opened 2023-12-08
Fix images link in en-ie

Pirionfr:
Fix images link in en-ie Getting started with Public Cloud Databases


OVH-Team-Guides:
Thank you for the fix @Pirionfr

PR #5526, opened 2023-12-11
Fix wording and add warning and information message

thild42:
Fix some wording and add messages for Advanced migration and disk size.


OVH-Team-Guides:

PR #5534, opened 2023-12-12
Fix typo in service account doc

yomovh:


OVH-Team-Guides:
Thank you for the fix @yomovh

PR #5582, opened 2023-12-20
cold archive overview update

fpillotovh:


pages/storage_and_backup/object_storage/cold_archive_overview/guide.de-de.md
```diff
@@ -1,9 +1,15 @@
...
 
+<style>
+.w-100 {
+    width:100% !important;
+}
+</style>
+
 ## Objective
 
 OVHcloud Cold Archive is  a class of storage designed for a long-term strage of non-actively used data.
```
lason-ovh: I would use "rarely accessed" data instead

pages/storage_and_backup/object_storage/cold_archive_overview/guide.de-de.md
```diff
@@ -39,13 +45,15 @@ The service allows you to focus on building and deploying cloud applications whi
 
 The service is fully managed by OVHcloud and accessible through S3 API.
 
-**3 steps operation**
+**5 steps operation**
 
-1. Store your data in an Object Storage bucket first in RBX Region
-2. Archive it on tapes
-3. Restore your data or/and delete your archive
+1. Store your data in an Object Storage bucket first in RBX Region – this bucket is a temporary bucket,
```
lason-ovh: region name is "rbx-archive", there is another region named "rbx" but is not the same

pages/storage_and_backup/object_storage/cold_archive_overview/guide.de-de.md
```diff
@@ -39,13 +45,15 @@ The service allows you to focus on building and deploying cloud applications whi
 
 The service is fully managed by OVHcloud and accessible through S3 API.
 
-**3 steps operation**
+**5 steps operation**
```
lason-ovh: '5 steps operation' is a bit misleading in my opinion.
It is not exactly one operation comprised of 5 steps but rather there are 3 possible operations : archiving, restoring, and deleting.

lason-ovh (changes requested):
the lifecycle.png is misleading: in the archiving flow, it says minimum_days=30 which is wrong, for the context, when I made the schema, it was to explain to @agoude the upcoming feature of being able to set a minimum retention days to remove any possibility to delete the archive within the aforementioned retention period.
Anyway, it we want to emphasize that there is a minimum engagement period for archives, we should at least update the value to 180 days

pages/storage_and_backup/object_storage/cold_archive_overview/guide.de-de.md
fpillotovh: done

pages/storage_and_backup/object_storage/cold_archive_overview/guide.de-de.md
agoude: Well, ok you're right. We should maybe use a "3-operation process: Store - Archive - Restore" instead, and explain there are other actions (move, delete..) also available. What do you think @lason-ovh?

pages/storage_and_backup/object_storage/cold_archive_overview/guide.de-de.md
agoude: agreed, and there is also a small typo on "strage" -> storage. Thanks

pages/storage_and_backup/object_storage/cold_archive_overview/guide.en-asia.md
```diff
@@ -39,13 +45,15 @@ The service allows you to focus on building and deploying cloud applications whi
 
 The service is fully managed by OVHcloud and accessible through S3 API.
 
-**3 steps operation**
+**5 steps operation**
```
agoude, line 48: As said before, we should indeed finally use a "3-operation process: Store - Archive - Restore" instead, and do the explanation for the optional ones (i'm going to comment the small changes).

pages/storage_and_backup/object_storage/cold_archive_overview/guide.de-de.md
```diff
@@ -39,13 +45,15 @@ The service allows you to focus on building and deploying cloud applications whi
...
 
-1. Store your data in an Object Storage bucket first in RBX Region
-2. Archive it on tapes
-3. Restore your data or/and delete your archive
+1. Store your data in an Object Storage bucket first in RBX Region – this bucket is a temporary bucket,
+2. Archive it on tapes,
+3. Retrieve your data
+4. Decide to move or suppress the data from the temporary bucket
```
agoude: for 4. and 5. could we try to mention there are post-archiving process (move/delete actions).

lason-ovh (changes requested):
the lifecycle schema has an error: in the archiving process, in step 2, it should be 180 instead of 30

lason-ovh (changes requested):
Text is good for me, except the lifecycle image that must be changed (see my last comment)

OVH-Team-Guides:
To be replaced by PR #6452

PR #5597, opened 2023-12-26
Update guide.fr-fr.md for MKS Load Balancer

antonin-a:
As we will release the usage of OVHcloud Load Balancer with MKS in Beta phase, we need to edit this doc to clarify the differences between LB for k8s and Network Load Balancer


OVH-Team-Guides:
Thank you for your contribution @antonin-a

PR #5645, opened 2024-01-10
FR add bouygues profil for internet

AxelDum:
* Add section to retrieve profil information from customer panel
* Add configuration section for bouygues provider


OVH-Team-Guides:

PR #5667, opened 2024-01-16
BrenKt-160124-OWA-Quickfix

BrenKt:


OVH-Team-Guides:
Actually, Outlook Web Access is a deprecated name.
OWA now stands for Outlook Web App
This PR is not valid. We should replace occurrences of Outlook Web Application with Outlook Web App.

PR #5672, opened 2024-01-16
fix(dedicated/partitioning_ovh): issue with removed paragraph

JayBeeDe:
Fix issue probably introduced in https://github.com/ovh/docs/commit/70cf961e69ea0c1ff096376de79886eee078442d
(not sure because too big commit so I cannot check)


JayBeeDe:
image

OVH-Team-Guides:
Thank you for the fix @JayBeeDe

PR #5700, opened 2024-01-22
Remove folder, files and update the index

thild42:
Data Integration product is being removed from the OVH API and the Manager.

This PR removes documentation and index links to Data Integration.


thild42:
The documentation was made after the migration so there was no redirection to remove. There is also no need to implement redirection for the missing documentation.

OVH-Team-Guides:

PR #5711, opened 2024-01-24
KevS-ScreenBankProject-Preparation-nomenclature-of-domainDNS-screens-1

Kazarus4:
screens correctly renamed for these guides : 

- dnssec
- dynhost
- dns server general infos
- dns zone create
- edit dns zone
- dns zone mx


Kazarus4:
old screens which was useless has been deleted too

PR #5745, opened 2024-01-31
S3 Object Storage - Getting started update -  Deletion of a bucket with versioning and object lock activated

lason-ovh:
added doc on deletion of a bucket with versioning and object lock activated


OVH-Team-Guides:

PR #5753, opened 2024-02-01
fix(dedicated/bring-your-own-image): update page according upcoming migration

JayBeeDe:
ref: PUBM-35205

I have also updated bring-your-own-linux to make bring-your-own-image and bring-your-own-linux as much consistent as possible


OVH-Team-Guides:

PR #5772, opened 2024-02-08
update Horizon guide with SSO between Manager & Horizon

yomovh:


OVH-Team-Guides:

PR #5819, opened 2024-02-21
KevS-SK1400-KFT0013797-Update-statistics-and-logs-for-WebHosting

Kazarus4:
Start a reorganization of this guide and add more precisions.

need personnal review for the last part Administration user, this about separate logs if necessary.


Kazarus4:
FRFR updated and reorganized.

More precisions added about manager graph too. 

Last function completed too.

Need proof by @Y0Coss

PR #5861, opened 2024-03-01
fix(ldp): update syslog_ng version

edefaria:
updating current ldp syslog-ng to use rfc5424 protocol from syslog-ng instead of setup by template.


OVH-Team-Guides:
Thank you for your contribution @edefaria

PR #5862, opened 2024-03-01
added new supported regions to high perf

lason-ovh:
added new supported regions to high perf


OVH-Team-Guides:

PR #5864, opened 2024-03-03
DNS Zone edition - Format improvements suggestions - JE emlyon

UsernameCharbelisnotavailable:


OVH-Team-Guides:
A compromise would be to use tabs instead of a table. This way, we would have an improved display and no orphan information boxes.

UsernameCharbelisnotavailable:
> A compromise would be to use tabs instead of a table. This way, we would have an improved display and no orphan information boxes.

Seems like a great idea ! Do you have an example of a guide where tabs where used? So I can align with OVHcloud´s guides?

OVH-Team-Guides:
You can find some examples on https://help.ovhcloud.com/csm/fr-dns-zone-dkim?id=kb_article_view&sysparm_article=KB0058101#configuration-complete-du-dkim
Here's the source: https://github.com/ovh/docs/blob/develop/pages/web_cloud/domains/dns_zone_dkim/guide.fr-fr.md?plain=1

PR #5873, opened 2024-03-05
New feature to keep local users available

gbarideau:
Updated screenshots with the new checkbox
Adding a warning about IAM policies for users with "NONE" role


OVH-Team-Guides:

PR #5971, opened 2024-03-19
Public Cloud DB - Logs to customers - how to find streamId

lpatte:
Following some feedback that the LDP stream ID is not easy to find.


bs-ovh (approved):

OVH-Team-Guides:
Thank you for your contribution @lpatte

PR #5991, opened 2024-03-20
Hk object storage s3 asynchronous replication

KebirHakim:
I've addressed the feedback and made the necessary corrections to our document regarding data synchronization and storage management strategies. The revisions reflect the current limitations and capabilities more accurately, especially concerning data replication across different storage classes


OVH-Team-Guides:
Replaced by https://github.com/ovh/docs/pull/6017

PR #6007, opened 2024-03-22
Regles_de_boite_de_reception_depuis_OWA_JEemlyon

Abla832:


BrenKt:
OWA interface for MXPplan is not related to your web hosting offer or a specific MXplan offer.  Roundcube is the current default interface for MXplan for the moment

PR #6047, opened 2024-03-27
Hb raci plateform.sh

H4516022Y:
hello team,

here is a new resonsibility model for devops pipeline using  plateform.sh

thanks for your review


OVH-Team-Guides:

PR #6093, opened 2024-04-04
Additional information -Configuring OVHcloud Secondary DNS on a VPS- JE EMLYON

UsernameCharbelisnotavailable:


tcpdumpfbacke:
Thank you for the contribution @UsernameCharbelisnotavailable ! We are closing this request. 
- Instructions should be focussed on the main KB topic and (in most cases) not repeat what is explained on other pages. This reduces maintenance cost for us. 
-This is not a beginner guide but aimed at users who admin their own DNS, so adding a TXT is not a task that needs detailed explanation (link to the DNS zone guide is available if further illustration of our Control Panel is required).

PR #6096, opened 2024-04-05
Public Cloud - Compute - apps_first_steps - JEemlyon

Abla832:


OVH-Team-Guides:
Thank you for your suggestion @Abla832 
However, the installation of WordPress is a custom one so it can not be used as a reference for this guide which is about **preinstalled** applications.
So We're closing this Pull Request.

PR #6100, opened 2024-04-05
additional-ip-buy_JEemlyon

Abla832:


OVH-Team-Guides:
Thank you for your suggestion @Abla832 
We consider this to be too obvious to be documented. In the Control Panel, the customer is advised to choose "a" country.
So we're closing this Pull Request

PR #6111, opened 2024-04-08
KevS-SK1491-Transfert-incoming-Gandi-TRANSLATIONS

Kazarus4:
All translations are done

Need proof by : 

- @tcpdumpfbacke for DE and EN subs
- @Y0Coss for latin and PL subs


OVH-Team-Guides (approved):
The bullet points list is broken in the primary for all translations and the FR version.
Aside from that, ES IT PL PT are ok

PR #6128, opened 2024-04-11
[cloudDatabases] adds tags on docs

ybriffa:


OVH-Team-Guides:
As seen with @BellionBastien, the tags need to be added inside the meta.yaml files.

PR #6131, opened 2024-04-11
Fix Nginx on MKS guide according to customer feedbacks

scraly:


OVH-Team-Guides:
Thank you for the fixes @scraly

PR #6149, opened 2024-04-15
Billing - Renewing services via API - Minor update - JE-EMLYON

camillebssa:


OVH-Team-Guides:
Hello @camillebssa 
Thanks for the suggestion but this guide is really intended for API usage, which is a specific use case for some of our customers. We don't think we need to add the information you suggested.
Best regards

PR #6159, opened 2024-04-15
Add encryption step on S3 bucket creation doc

agoude:
Add encryption step on S3 bucket creation doc:
- modify text
- add image


OVH-Team-Guides:

PR #6169, opened 2024-04-16
fix error in mongodb deploy with terraform documentation

vparpoil:


yomovh:
Thanks for pointing that. There was indeed an error !

yomovh:
Thanks for pointing that. There was indeed an error !

OVH-Team-Guides:
Thank you for your contribution @vparpoil

PR #6170, opened 2024-04-16
fix wrong user in mongoDB terraform article

vparpoil:
Following https://github.com/ovh/docs/pull/6169


yomovh:
Thanks for pointing that. There was indeed an error !

OVH-Team-Guides:
Thank you for your contribution @vparpoil

PR #6187, opened 2024-04-17
VCD Docs fix - naming / missing section

Hugo1380:
Change "vCD" to "VCD" according to VMware Cloud Director official doc: https://docs.vmware.com/en/VMware-Cloud-Director/index.html the official abbreviation is "VCD".

Fixed some doc lang that have `<a name="vDConOVH"></a>` instead of `<a name="VCDonOVH"></a>`.

Remove the "(OpenSource Only!)" mention in the VMware Marketplace comment, as there is some pre-packaged software available there that is not open source, but simply does not require a license.

Add missing "Virtual Machine Management" section for es-us lang.


OVH-Team-Guides:
Thanks very much for the updates @Hugo1380

PR #6200, opened 2024-04-18
fix(dedicated/api-os-installation): remove usespla example

JayBeeDe:
ref: PUBM-34717


OVH-Team-Guides:
Thank you for the fix @JayBeeDe

PR #6206, opened 2024-04-19
feat(dedicated/api-os-installation): add api call to share anchored link

JayBeeDe:
ref: PUBM-36345

We want to share https://help.ovhcloud.com/csm/en-dedicated-servers-api-os-installation?id=kb_article_view&sysparm_article=KB0061951#os-questions in terraform provider doc, but as you can see, customer arrives at this part of the page and doesn't know where this data comes from


OVH-Team-Guides:

PR #6215, opened 2024-04-22
Lso update s3 website hosting

lason-ovh:
additional step to set permissions on objects in bucket to set permissions


tcpdumpfbacke:
Thank you @lason-ovh !

PR #6277, opened 2024-05-08
Updated docs for MKS with vRack and Nodeport usage

mas-synaigy:
I added the information, that a MKS with vRack doesn't expose the Nodeport on public interfaces.


tcpdumpfbacke:
Thank you @mas-synaigy ! (Integrated into 6279)

PR #6320, opened 2024-05-16
SAP Logs - fix: syntax

remrozk:


OVH-Team-Guides:
Thanks for the fix @remrozk

PR #6333, opened 2024-05-20
update Public Cloud quotas to the current values

yomovh:
Based on VOC feedback.


OVH-Team-Guides:
Thanks for the fix @yomovh

PR #6359, opened 2024-05-24
Changing reference to user management menu

gbarideau:
Follwing MANAGER-12811, user guide have been updated to refer to the new user management menu

This update has to be sync with the manager update


OVH-Team-Guides:
To be edited as it also contains the KMS guides from #6283

PR #6388, opened 2024-05-29
fix(mks): update s3 client configuration in velero docs

mxm-tr:
The step to configure s3 credentials doesn't seem to appear in the current docs, so when trying to create the bucket the following error occurs:

```
$ aws --profile default s3 mb s3://velero-s3                                                                                
make_bucket failed: s3://velero-s3 Unable to locate credentials
```


OVH-Team-Guides:
Please do not use forks and update the `updated:` field for each version. Thanks

mxm-tr:
> Please do not use forks and update the `updated:` field for each version. Thanks

Ok, a new PR has been created thanks https://github.com/ovh/docs/pull/6391

PR #6406, opened 2024-06-04
S3 Object Storage - S3 location - target mapping date update

lason-ovh:
Typo correction


OVH-Team-Guides:
Thank you for the fix @lason-ovh

PR #6410, opened 2024-06-05
mks: add l4 and l40s as node flavors

nicolaslacroux:


OVH-Team-Guides:
Thank you for your contribution @nicolaslacroux

PR #6416, opened 2024-06-06
EN: Update guide.en-gb.md

Kieranwest:
Warn users about using 64bit Java with JViewer. In some instances attempting to Redirect CDROM or ISO causes JViewer to hard crash with no errors.


tcpdumpfbacke:
Thank you for your contribution @Kieranwest !

PR #6421, opened 2024-06-10
fix: syntax

remrozk:
🚀 Ready to prod


OVH-Team-Guides:
Thank you for the fix @remrozk

PR #6437, opened 2024-06-14
Correction on misstranlation in FR

gbarideau:
Some words were translated in FR and shoudn't be, and some weren't translated


OVH-Team-Guides (approved):

OVH-Team-Guides:

PR #6514, opened 2024-06-29
fix: change messages to _doc in OpenSearch API documentation

jehuty0shift:
This PR changes message to _doc which is the default value that must be used since our upgrade to OpenSearch. With the upcoming upgrade of OpenSearch 2.X, this guide must reflect the correct path to ensure we follow the guidelines of OpenSearch API.


OVH-Team-Guides:
Thank you for your contirbution @jehuty0shift

PR #6516, opened 2024-07-01
fix(ldp): clarify data gathering tools upgrade

pdepaepe:
Hello,

This will update the Logs Data Platform RACI, to explain how we update the data gathering tools:

- minors and patch are handled by LDP team
- majors are in the hand of the customer as they are a risk of incompatibility that only the customer can manage.

As always, thank you for your time on this.


OVH-Team-Guides:
Thank you for your contribution @pdepaepe

PR #6519, opened 2024-07-02
[S3 Object Storage] - S3 location - Updating regions table

lason-ovh:
update region urls table


OVH-Team-Guides:
Hello @lason-ovh , we should write guides in markdown whenever it is possible. HTML should be used for exceptions, when markdown formatting is showing too many constraints.
For this use case, we don't see any reason why the table can't be kept in markdown. Please revert to that format and also update the fr-fr version. Thank you.

OVH-Team-Guides (changes requested):
Please update the fr-fr version on top of the en-gb version.

OVH-Team-Guides (approved):

OVH-Team-Guides:

PR #6527, opened 2024-07-02
Fix AI S3 datastore add command

hyu9a:
Fix missing arg in ``ovhai datastore add``` command


OVH-Team-Guides:
Thanks for the fix @hyu9a

PR #6545, opened 2024-07-05
VMware - exporting logs to LDP - Minor fix on FR versions

OVH-Team-Guides:
FYI @pg-ovh


pg-ovh:
Thanks @OVH-Team-Guides

PR #6562, opened 2024-07-08
S3 Object Storage - Asynchronous replication - Table update

lason-ovh:
added precision about encrypted objects + update on table formatting because of the following feedback : "the table is a little misleading because at first glance, you get the impression that for each line, you have a correspondance between the left and right column, although the 2 columns are independant"


OVH-Team-Guides:

PR #6620, opened 2024-07-17
Public Cloud - Compute - Deploy a GPU instance - Updating availability information

mhurtrel:
Updated and precised info about GPU availbility


OVH-Team-Guides:
Thank you for your contribution @mhurtrel

PR #6642, opened 2024-07-19
fix(LDP): EN - rewrite Rust guide for US launch

jehuty0shift:
Hi this PR rewrites most of the code snippet used as example for the user. Indeed this guide has not been updated since 2023 and the snippets are no longer compatible with updated dependencies and the SDK writtent by OVH has also changed. 

The documentation change is necessary to validate the documentation before the US launch of LDP hence the Dependency Priority Label. 

Thank you !


bs-ovh (approved):

tcpdumpfbacke:
Thanky you @jehuty0shift !

PR #6647, opened 2024-07-22
feat(using_ipmi_on_dedicated_servers): update documentation

JayBeeDe:


pages/bare_metal_cloud/dedicated_servers/using_ipmi_on_dedicated_servers/guide.en-gb.md
```diff
@@ -23,65 +21,102 @@ Please note that you may also see the term KVM (Keyboard Video and Mouse) used f
 
 ## Instructions
 
-You can access the IPMI using several different methods, the Java applet (recommended), the KVM tool from a web browser (recommended but only available for the latest servers) or the browser (Serial over LAN).
+You can access the IPMI using several different methods¹. Here is a summary:
+
+|Method Name|Short Name|Description|Copy-Paste|Virtual ISO reader²|
```
JayBeeDe: @Y0Coss do you think it could be a good idea to add a column "Use Case examples" with example of use case for each of them ? Because SoL can be useful in case the server has booted but is not reachable over the network, because contrary to KVM you can copy-paste stuffs such as logs for example

bs-ovh: Thank you for the update @JayBeeDe ! I agree, please feel free to add any information you deem useful.

JayBeeDe: done  @tcpdumpfbacke

PR #6651, opened 2024-07-23
fix(LDP): port missing in Bonfire guide code snippets

jehuty0shift:
Hi, this PR fixes a missing port in this guide preventing users to connect the CLI to Logs Data Platform. The default port changed years ago but wasn't reflected in the guide. The US launch needs all LDP guides to be up to date, hence the dependency priority. 

After editing:
This PR also remove a non working code snippet from the guide and explain more another feature.

### DIFF
--- a/pages/manage_and_operate/observability/logs_data_platform/cli_bonfire/guide.de-de.md
+++ b/pages/manage_and_operate/observability/logs_data_platform/cli_bonfire/guide.de-de.md
@@ -1,6 +1,6 @@
 ---
 title: CLI - bonfire, querying graylog from a CLI
-updated: 2020-07-27
+updated: 2024-07-24
 ---
 
 ## Objective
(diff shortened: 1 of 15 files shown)


tcpdumpfbacke:
If it is a fix on an existing guide (looks like it), then I assume it is NOT an upcoming feature (Priority: Dependency)?

bs-ovh (approved):

tcpdumpfbacke:
Thank you @jehuty0shift !

jehuty0shift:
Thanks for the review, Indeed it is not an upcoming product. However this guide need to be updated to allow the proper launch of the LDP product in The US. They use OVH guides to populate their documentation website. If the guides are not working or obsolete, the launch won't happen. Hence the dependency. The launch is scheduled before the end of Q4 IIRC.

tcpdumpfbacke:
Thanks! For future reference: The label means that the guide update must wait for something else to happen. This is the opposite case. "Urgent" label would have been better. Or we need a new one. :)

PR #6692, opened 2024-08-06
Hpc - Migration to VCD tricky uses cases  guide update

pg-ovh:
Waiting for screenshot in EN.
EN alignement needed when FR is proof.


Y0Coss:
Replaced by https://github.com/ovh/docs/pull/6864

PR #6699, opened 2024-08-06
fix(LDP): EN - multiple guides update for US launch

jehuty0shift:
Hi, this Pull Request updates multiple guides, that's why i didn't duplicate the changes on all the languages this time to make the review easier. 
These guides contained typo or outdated configuration and commands that needs to be updated before the Launch of LDP in the US. 
I used the **Dependency Label** for this time since there will be multiple PR based on these changes to be made before this Thursday (2024/8/8). 
Thanks !


Y0Coss:

PR #6706, opened 2024-08-07
fix image

StevenSoulet:


StevenSoulet:
Modification to modify an image that includes the IP of the test account.

Y0Coss:
Thank you for the fix @StevenSoulet

PR #6720, opened 2024-08-08
VPS - Configure IPv6 - Netplan configuration example update

kristous:
netplan configuration example does not work with prefix /128 (gateway not in the same net)


Y0Coss (approved):

Y0Coss:
Thank you for your contribution @kristous

PR #6721, opened 2024-08-08
Update MKS Public Cloud loadbalancer guide to remove typo and Naming Convention

antonin-a:
A quick fix to remove typos and add explanations regarding the naming convention that needs to be respected if using an existing LB that has not been created by MKS


Y0Coss:
Thank you for your contribution @antonin-a

PR #6731, opened 2024-08-09
Lso update storage class naming

lason-ovh:
use official nomenclature for AWS storage classes to avoid confusion with customers


Y0Coss:

PR #6733, opened 2024-08-09
add ICMP filter on Load Balancer IP

yomovh:
Add the fact that the ICMP traffic is filtered on LB IPs


Y0Coss:

PR #6738, opened 2024-08-14
MKS - Add High Speed Gen2 as storage class

nicolaslacroux:


Jessica41:
thank you for your contribution @nicolaslacroux

PR #6767, opened 2024-08-21
 Hpc - Guide creation datastore upload on managed vSphere

pg-ovh:
Add fr-fr version for proofread.


Y0Coss:

PR #6785, opened 2024-08-27
Object Storage - S3 buucket with Terraform - Requirements update

lason-ovh:
add route in apiv6 to target S3


lason-ovh (commented):
ok pour moi

Y0Coss:

PR #6786, opened 2024-08-27
Hpc - Guide fix kms on managed vSphere

pg-ovh:
Add fix on all en versions.


Y0Coss:
Thank you for the fixes @pg-ovh

PR #6803, opened 2024-09-02
FB - EOL/EOS VPS distrib

fabienOVH:
As discussed, I replaced the existing content guide by the new one with all the distributions.


Y0Coss:

PR #6926, opened 2024-09-30
maj doc managed bare metal

StevenSoulet:
Hello,

Is it really necessary to keep the Managed Bare Metal documentation category?

It's a copy of the VMWare documentation, with the difference that it's not up to date.

Steven


Y0Coss:
We need to discuss with @franckcosta how to best handle this documentation.

Y0Coss:
Thank you for your contribution @StevenSoulet

PR #6933, opened 2024-10-01
Update S3 region endpoints guide.en-ie.md

agoude:
Adding SYD and MUM in the endpoints list


Y0Coss:

PR #6936, opened 2024-10-01
Local Zones for Compute - Capabilities update

EddyMedinaC:
Due to the new Instance backups feature to be released next week, the following information was updated:
Added:
| Instance Backups | Yes |
| Automatic Backups | No |

PR #6937, opened 2024-10-01
Update guide.en-ie.md

EddyMedinaC:
Due to the new Instance backups feature for local zones to be released next week, the following information was updated: 
| Instance Backups | Yes |
| Automatic Backups | No |


Y0Coss:
I'm closing this PR as everthing is already handled on #6936

PR #6977, opened 2024-10-04
Hpc - Guide fix migration VCD

pg-ovh:
Add guide fix fr + en version.


Y0Coss:

PR #6984, opened 2024-10-08
FB - Fix image on Multisite diagnostic

fabienOVH:


Y0Coss:
Thank you for the fix @fabienOVH

PR #6997, opened 2024-10-10
KevS-Optimisation-title-SSL-EV-guide-for-WebHosting

Kazarus4:
All title changed and translated

Need quick proof by : 

- @tcpdumpfbacke for DE and EN titles
- @Y0Coss or @BrenKt for latin and PL titles


bs-ovh (approved):

Y0Coss:
I'd be more in favour of `Web Hosting - title` rather than `Web Hosting: title`.
That would also be more coherent with the new naming chosen for Email guides.

PR #7017, opened 2024-10-14
Hpc - VCD Veeam guide update

pg-ovh:
Add fr + en version for guide update


Y0Coss:

PR #7029, opened 2024-10-16
Dedicated Servers - OLA on Debian 9 - Bond fix

Y0Coss:
@sbraz , please review the rest of the guide


pages/bare_metal_cloud/dedicated_servers/ola-enable-debian9/guide.en-asia.md
```diff
@@ -62,8 +62,8 @@ auto bond0
   bond-slaves eno1 eno2
```
sbraz, line 62: Can you deindent `iface bond0 inet static` please? It should not have leading whitespace.

I checked the rest of the configuration and I can confirm that the options work.

PR #7031, opened 2024-10-16
Lso update replication object lock

lason-ovh:
added infos on how to activate replication when there is object lock activated


Y0Coss:
Thank you for your contribution @lason-ovh .
Please make sure to edit the fr-fr and en-gb versions of guide, so to ease duplication of content on our side.

PR #7045, opened 2024-10-18
proxmox-network-HG-Scale: remove unneeded parameters + other fixes

sbraz:
* proxmox-network-HG-Scale: remove unnecessary sysctl parameters for vRack

  Remove ip_forward and proxy_arp for the vRack section. ip_forward is not
required because there is no IP forwarding done. The VMs are in a
network (the vRack) where they communicate directly with the gateway
(the gateway answers ARP requests in the vRack), so there is no need for
proxy_arp either. I tested this in SK-1686.

* proxmox-network-HG-Scale: set ip_forward in sysctl.conf for ADV

  Set ip_forward in sysctl.conf for the ADV range too. Setting it when
the interface comes up is not the best practice and we already fixed
this for HGR/SCALE ranges.

* proxmox-network-HG-Scale: remove references to Proxy ARP

  Proxy ARP is not required when the additional IP is routed either
because the VM communicates with the host via a specific 192.168.0.0/24
subnet. Its gateway is within this subnet, so the VM sends ARP requests
to 192.168.0.1 which is on the host, therefore Proxy ARP is not
required.

### DIFF
--- a/pages/bare_metal_cloud/dedicated_servers/proxmox-network-HG-Scale/guide.en-gb.md
+++ b/pages/bare_metal_cloud/dedicated_servers/proxmox-network-HG-Scale/guide.en-gb.md
@@ -363,8 +363,6 @@ iface bond0 inet dhcp
         bond-slaves ens33f0 ens33f1
         bond-miimon 100
         bond-mode 802.3ad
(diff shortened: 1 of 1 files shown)


Y0Coss:

PR #7051, opened 2024-10-18
Hpc - Guide update Zerto vCpu ram

pg-ovh:
Add guide update fr + en. Waiting for duplication.


Y0Coss:

PR #7054, opened 2024-10-19
Fix Kubernetes typo

dpilarsk:


Y0Coss:
Thank you for the fix @dpilarsk

PR #7060, opened 2024-10-21
Update - VCD migration use case

pg-ovh:
Add VCD migration use cases update guide : fr + en + duplication done.


Y0Coss:
Thank you for the fix @pg-ovh

PR #7066, opened 2024-10-22
Fix - Hpc VM encrypt with external KMS

pg-ovh:
Add guide fix (minor change).
En + fr version.


Y0Coss:
I don't see how this is a fix. In my opinion, the EN current version is OK.
FR current version is good too, we need singular (mot de passe) in French, not plural.

pg-ovh:
I have also removed an external link not needed. Proceed as you wish.

Y0Coss:
The `{.external}` doesn't break anything. So I prefer closing this PR. Thank you.

PR #7072, opened 2024-10-22
FR update overthebox version matrix & add information on dhcp renew

AxelDum:


Y0Coss:

PR #7105, opened 2024-10-29
Update - Hpc VMware connect iso

pg-ovh:
Add fr version for guide update.
Add info banner + go further ref to datastore upload guide.


Y0Coss:

PR #7126, opened 2024-11-04
Clarify 2024 server range

jslocinski:


Y0Coss:
Thank you for your contribution @jslocinski

PR #7130, opened 2024-11-05
Delete "renderer: networkd" from IP aliasing network configuration

GhaziTriki:
The `renderer: networkd` line was removed because, by default, Ubuntu uses `networkd` as the network renderer for managing network interfaces. Explicitly specifying `renderer: networkd` in the configuration is redundant in cases where the default network manager is already in use.

However, when a custom network manager, such as `NetworkManager`, is applied, defining multiple renderers can lead to conflicts. For example, if `NetworkManager` is managing the interface, specifying `networkd` as the renderer in this configuration can cause both network managers to attempt control, potentially resulting in unpredictable behaviour and network issues.

### DIFF
--- a/pages/bare_metal_cloud/dedicated_servers/network_ipaliasing/guide.de-de.md
+++ b/pages/bare_metal_cloud/dedicated_servers/network_ipaliasing/guide.de-de.md
@@ -281,7 +281,6 @@ Bearbeiten Sie anschließend die Datei mit folgendem Inhalt und ersetzen Sie `IN
 ```yaml
 network:
    version: 2
-   renderer: networkd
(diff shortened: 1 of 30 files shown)


sbraz:
Good point @GhaziTriki, merci :)
@Y0Coss, go ahead and merge this. I checked https://github.com/ovh/docs/blob/develop/pages/bare_metal_cloud/dedicated_servers/lacp-enable-netplan/guide.en-gb.md and we didn't add this line there.

sbraz (approved):

Y0Coss:
Thank you for your contribution @GhaziTriki

PR #7157, opened 2024-11-12
Compute - Create a custom image with Packer - Update with more recent software

sguyennet:
* Update version of Packer
* Add missing step to install the OpenStack plugin for Packer
* Update version of Ubuntu (16.04 is EOL)
* Use of an existing flavor (vps-ssd-1 doesn't exist anymore)
* Fix Packer variables (the build do not run otherwise)


Y0Coss:
Thank you for your contribution @sguyennet

PR #7166, opened 2024-11-14
Guide update - PCI - Test your disk speed - KFT0022677

Jessica41:


Jessica41:
Still have some reservation regarding the additional disk part, not sure we should document making the test on Linux. Checking with PU.

Y0Coss (approved):

PR #7171, opened 2024-11-15
Remove un-needed IPMI KVM step

DaMikeTheOne:
Removing IPMI KVM step as it is not required, and explain to set a password instead of a blank one : customer can directly connect to his server in RDP when a password is set which is simpler and more straight forward.

Removing it also adds support for older server with no IPMI KVM.


Y0Coss (approved):

Y0Coss:
Thank you for your contribution @DaMikeTheOne

PR #7204, opened 2024-11-25
Add new image - bucket creation FR-EN

agoude:
New image for the deployment mode selection - Step 2
following Nov 25th prod of S3 3-AZ


Y0Coss:
Thank you for the update @agoude 
Reminder: images mustn't exceed a 1400px width

PR #7206, opened 2024-11-26
Fix( 'configuring IPV6' documentation ): update the mistake of netplan command

OvhValentin:
There is an error in the Netplan command. We simply need to remove this error.

Refer to SK-1808


Y0Coss (approved):

Y0Coss:

PR #7208, opened 2024-11-26
fix(L3 services SNAT documentation): update the old version of the command

OvhValentin:
Refer to SK-1807.
PR to update the old version of the instance creation command. (In the past, this was a ‘s1-8’ flavour, but it no longer exists and has been replaced by a ‘d2-8’ flavour. )


Y0Coss (approved):

Y0Coss:

PR #7209, opened 2024-11-26
Add proper key auth section at SD reinstall

bs-ovh:


Y0Coss (approved):
All good but the date maybe (25 > 26)

PR #7210, opened 2024-11-26
LACP: remove hardcoded values for miimon and downdelay

sbraz:
See 7814cb15d49d1f for more details.


sbraz:
@Y0Coss I think the CentOS 7 guide could be removed at some point as it's been EOL for 6 months now.
This PR also touches Nutanix stuff but it's in a scenario where we replace a Nutanix VM with a dedicated server so I think it's in our scope. Pinging @malacan for visibility.

Y0Coss (approved):

Y0Coss:
Thank you for the update @sbraz

PR #7215, opened 2024-11-27
FR/EN Add Hold an automatic EFS snapshot

dariakts:


Y0Coss:
Translations will be generated in a 2nd PR

PR #7230, opened 2024-12-02
proxmox-network: fix indent of ifupdown example

sbraz:
@Y0Coss no need to update the last modification date. It's just cosmetic but it bothered me :)


Y0Coss (approved):

Y0Coss:
Thank you for the fix @sbraz

PR #7251, opened 2024-12-05
Vcd migration use cases - fixing mistake

dstagnol:
Fixing mistake in Post Action Table (in section Step 2 - After migration (mandatory))


Y0Coss:
Duplicated content

PR #7296, opened 2024-12-13
update guide and center imgs

MathieuBsqt:
Back in the time, we used to only have the `AI Training` product. This is why this guide was only mentioning this product, and not all the existing AI Solutions (`AI Notebooks`, `AI Training` & `AI Deploy`). 

This update is there to fix that. However, note that the role is still named after AI Training.

Note that I've:

- tried to center the images using style.
- modified guide excerpt


Y0Coss (approved):

Y0Coss:

PR #7297, opened 2024-12-13
Update doc about VMAC

Stoufiler:
Since deployment of vmac on ADV Gen3 
We need to update the documentation to inform customer that only 32 VMACs are available on these servers


Y0Coss:
Replaced by #7298

PR #7298, opened 2024-12-13
Update vmac documentation

Stoufiler:
Since deployment of vmac on ADV Gen 3 servers yesterday

We should inform customers that only 32 vMACs are allowed on this range of servers


Y0Coss:
Thank you for your contribution @Stoufiler

PR #7301, opened 2024-12-15
DNS DynHost - PL translation fix

saikotek:
Fix auto-translated unintelligible sentence


Y0Coss:
Thank you for your contribution @saikotek.
We indeed use Systran engines to translate our source pages (French or English) and do our best to fine-tune delivered translations. We apologise for any bad translation you may find on our documentation.

PR #7302, opened 2024-12-16
VMware Cloud Director - Updating migration dates

dstagnol:
Updating dates on the EN-GB and FR-FR guides.


Y0Coss:

PR #7318, opened 2024-12-18
[AI & ML] - S3 compatible fix

MathieuBsqt:
- Guides were updated (only `gb` version). 
- Screenshots were updated.
- S3 video was removed. I've asked for its deletion on YT.
- First S3 term has been added a '*' for each guide.
- Footer S3* definition was added to each guide.

The only things I'm not sure about:
- Some CLI commands include "s3" parameter. I have not changed them. Otherwise, these commands won't work anymore.
- I've renamed a guide that includes "S3" in its title. As this is a title, I haven't used the '*' symbol. It only appears in the excerpt.


Y0Coss:
Of course, `s3` commands need to remain unchanged.
The new title is OK. Thanks a lot @MathieuBsqt 
I'll check everything and duplicate on due time.

@agoude FYI

Y0Coss (approved):

Y0Coss:

PR #7351, opened 2024-12-30
S3 compatible compliance update

MathieuBsqt:
Here is an update of the S3 documentation to: 

- Detail whole process
- Adapt it to 1 AZ, 3 AZ and local zone new feature
- Provide further support for someone who has not yet created an Object Storage bucket (interface is a bit different)

Images were resized and "S3 compatible" term has been used.


Y0Coss (approved):

Y0Coss:

PR #7356, opened 2025-01-02
EN-LDP related-minor fixes

SR-OVH-US:
This branch was started months ago and I'm just now revisiting and finishing it up. I hope the time gap doesn't cause problems.


Y0Coss:
Thank you for the fixes @SR-OVH-US

PR #7357, opened 2025-01-03
Update available Kafka version

thild42:
Remove depreciated version and add new available versions.


Y0Coss:

PR #7364, opened 2025-01-06
Properly hide temporary credentials

thild42:
Hide temporary credentials in the screenshots


Y0Coss:
Thank you for the fix @thild42

PR #7389, opened 2025-01-10
MKS - Backing up a cluster with Velero - Updating the endpoints

francoisovh:
correcting the S3 endpoints


Y0Coss:
As agreed with you @francoisovh , this needs to be re-done. Thank you for your contribution.

PR #7404, opened 2025-01-14
Feat(databse refacto): create troubleshooting guide

OvhValentin:


Y0Coss:
The index needs to be edited

PR #7406, opened 2025-01-14
Feat(database refacto) : create restore backup guide

OvhValentin:


Y0Coss:
Index edition is needed

PR #7407, opened 2025-01-14
fix(general guide deployment): update def

OvhValentin:


OvhValentin:
Need to add index

Y0Coss:

PR #7416, opened 2025-01-15
adding on-prem mention for Netapp and Barbican compatibilty

gbarideau:


Y0Coss:

PR #7457, opened 2025-01-28
Managed Private Registry improvments

thcdrt:


XavierDuthil (approved):

cortexx (approved):

nicolaslacroux (approved):

Y0Coss (changes requested):
Hello @thcdrt , as you remove a guide, you also need to delete the corresponding line in the index file:
`+ [End Of Life (EOL) Harbor 1.x - How to replicate and migrate your data to Harbor 2.x](public_cloud/containers_orchestration/managed_private_registry/eol-harbor-1.x)`

Should we also write 301 rules to redirect the URL? In this case, what should be the target of the 301?

thcdrt:
Removed from index, thanks @Y0Coss 
About the redirection I don't know, I think it's ok to not have one.
WDYT @nicolaslacroux  ?

Y0Coss (approved):

Y0Coss:
Thank you for your contribution @thcdrt 
I can write redirections later if you think it's necessary.

nicolaslacroux:
Hello @Y0Coss 

I share the same opinion that Thomas, I don't see in which case the redirection would be relevant here, there is no call from the manager or other guide.
it is good to know for the future that there is this redirection feature

PR #7460, opened 2025-01-28
[PUBM-41432] pxe-with-full-private-dedicated: remove exception for T3

sbraz:
T3 servers now have working fallback when LACP is not enabled, allowing them to use DHCP to boot.


Y0Coss:

PR #7291, opened 2024-12-12
Update guide.en-ie: Linux CIFS Backup mount parameters

anttiryt:
sec=ntlm removed from Linux kernel so removed from mount instructions.

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=76a3c92ec9e0668e4cd0e9ff1782eb68f61a179c

Mounting Cifs by hostname on Ubuntu requires additional package cifs-utils

https://serverfault.com/questions/609365/cifs-mount-in-fstab-succeeds-on-ip-fails-on-hostname-written-in-etc-hosts

PR #7485, opened 2025-01-31
Update guide.en-ie: Linux CIFS Backup mount parameters

sbraz:
Based on #7291. @anttiryt, are you OK with this version?

sec=ntlm removed from Linux kernel so removed from mount instructions. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=76a3c92ec9e0668e4cd0e9ff1782eb68f61a179c

Not specifying the version results in:
```
Dialect not supported by server. Consider  specifying vers=1.0 or vers=2.0 on mount for accessing older servers
```

Mounting Cifs by hostname on Ubuntu requires additional package cifs-utils

https://serverfault.com/questions/609365/cifs-mount-in-fstab-succeeds-on-ip-fails-on-hostname-written-in-etc-hosts


bs-ovh (approved):

Y0Coss:
Thank you for your contribution @anttiryt and @sbraz 
I will also close #7291

PR #7488, opened 2025-02-03
fix(analytics): add images for terraform init

OvhValentin:


Y0Coss:

PR #7495, opened 2025-02-04
S3 - FAQ - Offsite Replication option details

agoude:
Add in FR and EN-GB details about the Offsite Replication option (released on Feb 3rd)


Y0Coss:

PR #7496, opened 2025-02-04
Help Center update - Switching to 16 Knowledge Bases

Y0Coss:
The current version of the Help Center displays 9 Knowledge Bases.
This Pull Request contains:

- An updated index for 16 Knowledge Bases
- Translations for the updated/created new index categories
- Updated reference categories 
- Deletion of the Labs Knowledge Base which now only contains EoL products


Y0Coss:
This has to revamped before considering a merge.

PR #7511, opened 2025-02-07
Added APNIC support

SlimJ4:
The APNIC registrar will be added to the list of supported registrars shortly - modifications were requested to reflect that.


Y0Coss:

PR #7516, opened 2025-02-10
BYOIP - FAQ edit after feedback

SlimJ4:
Feedback provided by the PM points to obsolete FAQs. One was modified to use broader terms, so that APNIC is included by default. Another was deleted because of redundancy.


Y0Coss:
Thnak you for your contribution @SlimJ4

PR #7536, opened 2025-02-13
fix(savings plan creation): add images and fix

OvhValentin:


Y0Coss:

PR #7538, opened 2025-02-13
fix(savings plan): fix space and img

OvhValentin:


Y0Coss:

PR #7557, opened 2025-02-19
BYOIP - APNIC support update

SlimJ4:
This version aggregates all the previous edits that were rolled back due to an early merge, and acknowledges current experimental support of NIRs instead of stating they are not supported at all.

ETA of the feature: 2025/02/26


Y0Coss:

PR #7559, opened 2025-02-19
Updating guide.en-us.md start instance on attached volume

fabver5:
old openstack commands regarding cinder and nova
no mention that the rescue pro is not available with a bootable volume.
add volume creation in horizon and cli since no in the docs
add horizon steps with tab to separate with cli
remove configuring the volume and hard reboot for the instance since now done automatically if volume is created prior to the instance


Y0Coss:
Replaced by #7560

PR #7560, opened 2025-02-20
Update start instance on attached volume

fabver5:
Update guide.en-us.md
old openstack commands regarding cinder and nova
no mention that the rescue pro is not available with a bootable volume.
add volume creation in horizon and cli since no in the docs
add horizon steps with tab to separate with cli
remove configuring the volume and hard reboot for the instance since now done automatically if volume is created prior to the instance

This PR replaces #7559


Y0Coss:
Thank you for your contribution @fabver5

PR #7562, opened 2025-02-20
chore(mks) : update version 1.32

atnbrg:


Y0Coss:
Thank you for your contribution @atnbrg

PR #7591, opened 2025-02-27
Update proxy link in Kubernetes Dashboard intall guide

salquier:
Since Kubernetes Dashboard 7.0.0, nginx has been replaced by kong, so the way to access the dashboard has changed


Lucasgranet:
Hello @salquier 
Thank for you PR.

As mentioned on Kubernetes/Dashboard repo:

> As of version 7.0.0, we have dropped support for Manifest-based installation. Only Helm-based installation is supported now. Due to multi-container setup and hard dependency on Kong gateway API proxy it would not be feasible to easily support Manifest-based installation.

We need to update the documentation to switch for a Helm-based deployment (instead of a manifest one).

Since the documentation is deploying an old version of the Dashboard, the proxy link is still working.

Y0Coss:
Hello @salquier
As mentioned by @Lucasgranet , this guide will require a more complete update. Our MKS product team will handle it.
Thank you for your Pull Request.

PR #7593, opened 2025-02-27
BYOIP - Campus to Regions + WHOIS Record

SlimJ4:
Edits requested by the product team:
- Replace all mentions of campuses by regions
- Update the regions list => Replaced the list with a link to the product page for ease of maintenance, as suggested
- Added a precision about how the desired IP range must match the WHOIS record for the IP block


Y0Coss:

PR #7627, opened 2025-03-07
[EN-GB]-VCD-Network concepts

SR-OVH-US:
Deleted duplicate information in Step 2 (types of supported networks were shown in a table and tabs).

Renumbered Limitations as Step 3 (rather than Step 4)


Y0Coss (approved):

Y0Coss:
Thank you for the fix @SR-OVH-US

PR #7632, opened 2025-03-10
SAP - fix: typo

remrozk:


Y0Coss:
Thank you for the fix @remrozk

PR #7654, opened 2025-03-18
[PNRv2] WebHosting & Domain/DNS

Kazarus4:
KevS-PNRv2-SK1920-DomainDNS-and-WebHosting-finalisations

DynHost fully prepared.

All concerned guides are prepared too (except screens) in FRFR version.

Need proof for FRFR (only sentences which was updated) before FRCA duplications and translations for all subs.

PR #7655, opened 2025-03-18
[PNRv2] WebHosting & Domain/DNS

Kazarus4:
KevS-PNRv2-SK1920-DomainDNS-and-WebHosting-finalisations

DynHost fully prepared.

All concerned guides are prepared too (except screens) in FRFR version.

Need proof for FRFR (only sentences which was updated) before FRCA duplications and translations for all subs.


Y0Coss:
Error duplicate of https://github.com/ovh/docs/pull/7654

PR #7661, opened 2025-03-20
Ag object user mgmt

agoude:
- adding new actions in the supported action list


lason-ovh (approved):
ok pour moi

Y0Coss:
Should this be pushed to prod on March 24th (not before)?

agoude:
it can be prodded before, i didn't want to preempt any other tasks ;) thanks!

Y0Coss:

PR #7671, opened 2025-03-24
fix(3AZ ref archietcture): update array

OvhValentin:


Y0Coss:

PR #7683, opened 2025-03-26
BGP diagrams + clarity update

SlimJ4:
Diagrams were updated following client feedback.
A table was updated to make it clearer by using one term for IP prefixes instead of four different terms interchangeably.


Y0Coss:

PR #7841, opened 2025-05-09
Object Storage - Update endpoint mapping

lason-ovh:

- Update of existing guide(s)

## Description
This in an intermediary update in preparation of future GA of new S3 Infrequent Access tier. The objective is to provide customers with a head start regarding incoming changes to AWS/OVH storage class mapping.

## Mandatory information

- This Pull Request should be merged asap AND after review from @agoude

- This Pull Request content should be replicated for the US OVHcloud documentation : YES except the part regarding the geo-availability table containing the different endpoints by region because the US has their own endpoints

### DIFF
--- a/pages/storage_and_backup/object_storage/s3_location/guide.de-de.md
+++ b/pages/storage_and_backup/object_storage/s3_location/guide.de-de.md
@@ -1,6 +1,6 @@
 ---
 title: Object Storage - Endpunkte und Objektspeicher mit Geo-Verfügbarkeit (EN)
-updated: 2024-11-25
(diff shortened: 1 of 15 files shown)


agoude:
that's ok on my side! thx @lason-ovh

Y0Coss:
@lason-ovh & @agoude I've improved the footnotes formats so that it's easier for customers to retrieve the right information from each note (there are 4 different footnotes).
Please let me know if this is OK before I merge it.

lason-ovh:
ok for me

Y0Coss:

PR #7932, opened 2025-05-30
fix(bare_metal_cloud/partitioning_ovh): update known errors

JayBeeDe:

- fix

## Description

Update "Common customer errors" for dedicated server partitioning

ref: PUBM-42758

## Mandatory information

- This Pull Request can be merged as soon as possible when @sbraz has acked it


pages/bare_metal_cloud/dedicated_servers/partitioning_ovh/guide.en-gb.md
```diff
@@ -538,16 +538,15 @@ The following table gives an overview of well known customer errors and how to f
...
+|`Debian` does not provide a precompiled `ZFS` module and this server does not have enough memory to compile it (8 GB minimum required). Please use a different file system.|We always try to find the best balance between integration in OVHcloud ecosystem, hardware compatibility and reduced OS image size while providing an image that is as similar as possible as the official cloud-ready image provided by the software editor. `ZFS` module is not natively provided by the software editor, we don't want to make the image heavier for customer not needing `ZFS` and we want to offer a `ZFS`-ready OS with configured `ZFS` partitioning for customers who want it. Therefore `ZFS` module is compiled and installed during the OS reinstallation only for customers who ask for it. Since compilation requires memory in addition to that already used to run the rescue OS in memory, we have set the threshold at 8GiB to guarantee smooth operation and customer experience|- Remove all `ZFS` partitions or change their filesystems to something different than `ZFS`<br />- Choose another OS that natively supports `ZFS` filesystem such as Proxmox<br />- Use a server with 8GiB memory or more|
```
Y0Coss:
```suggestion
|`Debian` does not provide a precompiled `ZFS` module and this server does not have enough memory to compile it (8 GB minimum required). Please use a different file system.|We always try to find the best balance between integration in OVHcloud ecosystem, hardware compatibility and reduced OS image size while providing an image that is as similar as possible to the official cloud-ready image provided by the software editor. The `ZFS` module is not natively provided by the software editor. We don't want to make the image heavier for customers not needing `ZFS` and we want to offer a `ZFS`-ready OS with configured `ZFS` partitioning for customers who want it. Therefore the `ZFS` module is compiled and installed during the OS reinstallation only for customers who ask for it. Since compilation requires memory in addition to that already used to run the rescue OS in memory, we have set the threshold at 8GiB to guarantee smooth operation and customer experience|- Remove all `ZFS` partitions or change their filesystems to something different than `ZFS`<br />- Choose another OS that natively supports `ZFS` filesystem such as Proxmox<br />- Use a server with 8GiB memory or more|
```

Y0Coss (commented):
Minor EN text edits

pages/bare_metal_cloud/dedicated_servers/partitioning_ovh/guide.de-de.md
```diff
@@ -538,16 +538,14 @@ The following table gives an overview of well known customer errors and how to f
...
...e same zpool||`Debian` does not provide a precompiled `ZFS` module and this server does not have enough memory to compile it (8 GB minimum required). Please use a different file system.|We always try to find the best balance between integration in OVHcloud ecosystem, hardware compatibility and reduced OS image size while providing an image that is as similar as possible to the official cloud-ready image provided by the software editor. The `ZFS` module is not natively provided by the software editor. We don't want to make the image heavier for customers not needing `ZFS` and we want to offer a `ZFS`-ready OS with configured `ZFS` partitioning for customers who want it. Therefore the `ZFS` module is compiled and installed during the OS reinstallation only for customers who ask for it. Since compilation requires memory in addition to that already used to run the rescue OS in memory, we have set the threshold at 8GiB to guarantee smooth operation and customer experience|- Remove all `ZFS` partitions or change their filesystems to something different than `ZFS`<br />- Choose another OS that natively supports `ZFS` filesystem such as Proxmox<br />- Use a server with 8GiB memory or more|
```
sbraz: Looks like you accidentally merged the zpool name and the ZFS module lines

sbraz: I thought this was part of the PR so I reviewed it and I think the following would be slightly easier to read:
```suggestion
|`ZFS` partition already exists with zpool name `n`. Either choose another name for the `m` partition or set the same RAID level for all partitions within zpool `n`|Assigning multiple `ZFS` partitions to the same zpool name is possible: it means each dataset will be part of the same zpool. This is possible only if all the datasets (therefore the partitions defined via the API) have the same RAID level|- Set the same RAID level on all `ZFS` partitions which will be part of the same zpool<br />- Choose another zpool name: this partition will not be part of the same zpool<br />- Don't provide any zpool name: we will assign a default zpool name and this partition will not be part of the same zpool|  
```

sbraz: Why no leading hyphen in the second column? All other rows seem to have one.

pages/bare_metal_cloud/dedicated_servers/partitioning_ovh/guide.de-de.md
```diff
@@ -538,16 +538,14 @@ The following table gives an overview of well known customer errors and how to f
...
+|This operating system doesn't support `mdadm` metadata version 1.2 for `/` partition. Fallback to `mdadm` version 0.90 is not possible with `/` partition that requires `x` MiB on every disk, which exceed 4TiB on every disk.|- You have configured a `/` partition on top of a software RAID with an operating system that has a specific Linux kernel that only supports `mdadm` metadata version 0.9 for the `/` partition. Unfortunately this version of `mdadm` metadata doesn't work when the size used per disk is bigger than 4TiB|- Reduce the size of this partition so that it doesn't use more that 4TiB per disk<br />- Disable software raid, by restricting OS installation to 1 single disk<br />- Choose another operating system|
```
sbraz:
```suggestion
|This operating system doesn't support `mdadm` metadata version 1.2 for `/` partition. Fallback to `mdadm` version 0.90 is not possible with `/` partition that requires `x` MiB on every disk, which exceed 4TiB on every disk.|You have configured a `/` partition on top of a software RAID with an operating system that only supports `md` metadata version 0.9 for the `/` partition. Unfortunately this version of `md` metadata doesn't work when the size used per disk is larger than 4 TiB|- Reduce the size of this partition so that it doesn't use more that 4 TiB per disk<br />- Disable software raid, by restricting OS installation to 1 single disk<br />- Choose another operating system|
```

This is not a kernel restriction, all Linux kernels support metadata 1.2. The issue is lack of initramfs. I don't think we need to go into specifics.

pages/bare_metal_cloud/dedicated_servers/partitioning_ovh/guide.de-de.md
JayBeeDe: > Looks like you accidentally merged the zpool name and the ZFS module lines

No it's removed because that check is now part of the API, API errors are not going to be documented here

sbraz (approved):

sbraz:
@Y0Coss on te laisse éditer les colonnes "détails" et "solutions" si besoin, on a terminé les modifs.

Y0Coss:

PR #8032, opened 2025-06-19
Remove the style from the advanced parameters guides

thild42:
Remove the specific style that mess with the default style.


Y0Coss:
Thanks for the fix @thild42

PR #8033, opened 2025-06-19
[AI Endpoints] - replace decomisionned model

MathieuBsqt:
Small update to replace a model that has been decomisionned on our Catalog.


Y0Coss:

Issue #8038, opened 2025-06-19
question about optimising the site and the documentation procedure


DoStupidThingsFasterWithMoreEnergy:
@Y0Coss 

Hello

Why not use github hooks like pre-commit and or pre-merge to check the links / images and refuse the action if the hook output is not ok?

And depending on your hook, you can ensure that an irregular contributor will only be filtered by the pre-merge hook, so it's you who will receive the alert, whereas a regular contributor who knows the mechanisms will be filtered by the pre-commit hook and will even be warned before being able to commit. To do this, all you need to do is draw up a list of trusted contributors and, of course, make the hook accordingly.

Have you already run Selenium or Cypress type tests (for example) on your documentation site? You regularly have to refresh the page to access the kb ‘categories’.

This would indicate everything that is not :
- missing elements
- link error
- image loading error or to slow
- slow execution
- ...

Another question: why resize an image to 1400px max width? Personally I have a very large screen resolution (LG 22MD4KA-B UltraFine 4K Display 21.5" UHD 4096x2304) so when I click on an image it is smaller. Why not use the html picture tag and its source & srcset argument & default value?

example:

``` html
<picture>
 <source media="(min-width:650px)" srcset="img_pink_flowers.jpg">
 <source media="(min-width:465px)" srcset="img_white_flower.jpg">
 <img src="img_orange_flowers.jpg" alt="Flowers" style="width:auto;">
</picture>
```

*Example: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_picture just change the size of your browser*

This allows you to define minimum/maximum widths or minimum/maximum heights and, depending on the device, load only the correct image.

There's a big difference between a smartphone/tablet and a computer.

Many markdown to html systems have this plugin

And you could only apply it in production by creating images resized according to your breakpoint css and thus keep your images in your documentation whatever their size.

PR #8070, opened 2025-06-24
Fix global links in KB Public cloud

DoStupidThingsFasterWithMoreEnergy:
The purpose of these pull requests (Fix global links in KB ...) is to replace all links referenced in /links with their `/links` values.

*Example: replace all `https://www.ovh.com/auth/?action=gotomanager&from=https://www.ovh.co.uk/&ovhSubsidiary=GB` links with `/links/manager`*.

I've seen that a lot have been replaced by your work, but I've also found a lot, which is why I'm making pr by kb.

These are not a priority, however, I think they can help you make the documentation source more consistent and can prevent possible errors.

Sorry for closing my last pull requests, but I noticed a first error followed by a second one... so I fixed them and for me it's now OK.

FYI it is also possible to automate it using the methods I mentioned in the issue: https://github.com/ovh/docs/issues/8038

Be free to change the titles or to ask me any question.

Thank you for your reviews.

@Y0Coss no need for the `Fix` label, maybe `Minor Update` and even then, I'll let you be the judge.

> [!NOTE]
> **Sorry I know the 3000 views limit on github so if need tell me for this PR .... 3 560 is a bit too much...**
> Total time 00:02:24
>
> It's the longest time I've seen for a kb and the one that contains the most guides but with a lot of duplication during this processing... the time announced before only concerns this kb but all the languages and subs... so in a pre-commit, pre-merge hook or a github action in an automated way for each commit or merge... try to think about it


Y0Coss:
Indeed I would recommend cancelling this pull request and splitting the work into several smaller ones.
We (the OVHcloud documentation team) need to be able to assess each and every change.

I agree we need to put some automation to work here, we're considering it seriously. 

We are also thinking about reducing the number of duplicated versions now that we have a language fallback mechanism in place. This will take us some time but we will be working on it.

DoStupidThingsFasterWithMoreEnergy:
How many splits would be best? around 200? around 500? around 1000?

before closing the PR, what is the best option?

Y0Coss:
Around 1000 files in each PR should be manageable on our side and probably less work on yours. Thanks!

DoStupidThingsFasterWithMoreEnergy:
close before rewrite without `<>`  links

DoStupidThingsFasterWithMoreEnergy:
I'm waiting for min 19h or min 21h to update...maybe later... sorry...

PR #8083, opened 2025-06-26
[AI Guides & Tutorials] - Fix docker commands

MathieuBsqt:
Include both docker build and docker buildx command to make sure users can build an image that is compliant with OVHcloud AI Products.


Y0Coss:

PR #8119, opened 2025-07-02
Fix ingress-nginx acme tls challenge

scraly:

- Fix

## Description

The new version of iingress-nginx have blocking changes so without this modifications our customers will have troubles on their new certificates.

## Mandatory information


Jessica41:

PR #8122, opened 2025-07-03
EN/FR Manage EFS with Terraform

dariakts:

- New guide(s) 

## Description

Guide that describes how to use OVHcloud Terraform provider newest version that includes EFS resources and data sources with an EFS service.

## Mandatory information


Y0Coss:

PR #8131, opened 2025-07-04
Lso patch replication doc

lason-ovh:

- Update of existing guide(s)

## Description

Token is no longer needed to use replication in buckets with object lock.

## Mandatory information

- This Pull Request shouldn't be merged before: 2025-07-10 


Y0Coss:

PR #8150, opened 2025-07-10
IPLB Articles global update 1

SlimJ4:

- Update of existing guide(s)

## Description

Update to the outdated guides pertaining to the OVHcloud Load Balancer. Mostly updates to naming conventions.
Will need to be updated again as soon as all elements on the OVHcloud Control Panel are uniformized (several uses of "Failover IP" or "Datacenter" are remaining, for example).

## Mandatory information


Y0Coss:
@SlimJ4 , can you please update the fr-fr and fr-ca files as well?

SlimJ4:
@Y0Coss , will do ! Since there are differences in the naming conventions used in the FR and EN interfaces, some updates may be different or not required yet.

PR #8204, opened 2025-07-22
[VMware] Changing port for RSA endpoint and updating KMS regions

gbarideau:

- Update of existing guide(s)

## Description

Various update on this documentation following a support case : 
- Changing port to use for new RSA endpoint
- Updating available KMS region
- Removing beta mention

## Mandatory information


Y0Coss:

PR #8265, opened 2025-08-08
fix: public-cloud-terraform: remove provider/alias informations from examples

rbeuque74:

- Update of existing guide(s)

## Description

alias and provider keyword are an expert way of doing Terraform, and is complexifying the quick start example for no added value.

Customers that add resource must take this complexity into account and can lead to complicated errors.
Removing the complexity is better, as advanced users that want to deal with multiple configuration of the same provider know how to do that.

## Mandatory information


Jessica41:
Thanks! @rbeuque74

PR #8269, opened 2025-08-11
Object Storage - add support for whitelist ips

lason-ovh:

- Update of existing guide(s)

## Description

We recently added suppport in user policies for conditions based on ip addresses i.e we can allow/deny permissions based on the requester's ip address.

## Mandatory information

- This Pull Request shouldn't be merged before notification from @lason-ovh because it is still currently in QA


Y0Coss:

PR #8285, opened 2025-08-18
fix: mongoDB behaviour when the disk is full

fallmor:

- Fix
- Optimization

## Description

When the mongoDB  disk is full, all writes operations are blocked for all users except for the Admin.

## Mandatory information


Y0Coss:
Thank you for your contribution @fallmor

PR #8351, opened 2025-09-04
Update VoIP guide on service activation/deactivation

adrien-martin:
## What type of Pull Request is this?

- Update of existing guide(s)
- Fix

## Description

This PR aims to change the documentation about service activation/deactivation to match the equipment configuration.
Phone locking has no switch feature, the one set in this column (`*#33#`) is also used to query the feature state.

Also the code used to lock the phone (`*33*(CODE)#`) can be miss-leading, one might think this is a way to define the pin code, this code already configured.
So the PR adds a comment about default pin code and how to define it.

## Mandatory information

- This Pull Request can be merged as soon as possible.

- This Pull Request content should be replicated for the US OVHcloud documentation : NO

PR #8360, opened 2025-09-04
Update VoIP guide on service activation/deactivation

adrien-martin:

- Update of existing guide(s)
- Fix

## Description

This PR aims to change the documentation about service activation/deactivation to match the equipment configuration.
Phone locking has no switch feature, the one set in this column (`*#33#`) is also used to query the feature state.

Also the code used to lock the phone (`*33*(CODE)#`) can be miss-leading, one might think this is a way to define the pin code, this code already configured.
So the PR adds a comment about default pin code and how to define it.

This PR replace and update https://github.com/ovh/docs/pull/8351

## Mandatory information

- This Pull Request content should be replicated for the US OVHcloud documentation : NO


Y0Coss:
Thank you for your contribution @adrien-martin

PR #8377, opened 2025-09-09
[AI] - remote connection - img fix

MathieuBsqt:
- Fix


Y0Coss:
Thanks for the fix @MathieuBsqt

PR #8422, opened 2025-09-22
KevS-SK2348-Creation-of-the-Domain-and-DNS-FAQ

Kazarus4:
FRFR done.

Need peer and tech rv before pursuive with translations.

(We need to adapt 3 topics about "retrieve admin, tech and billing contacts for a domain name" because the procedure exist only for FRFR (or FR subs).)


Kazarus4:
All translations are done. 

Need proof by : 

- @tcpdumpfbacke for DE and EN subs.
- @benchbzh for latin and PL subs

Y0Coss (approved):

PR #8459, opened 2025-09-30
[PUBM-48237] BYOLinux: add more details to image boot mode specs

sbraz:

- Update of existing guide(s)

## Description

* [PUBM-48237] BYOLinux: add more details to image boot mode specs

    If we want to install a Linux distribution on a legacy boot server, our
    image should support it, the same goes for UEFI.
    Also, we don't really need to mention GPT/MBR as we create the
    partitions ourselves, we only mount the QCOW2 image and extract its
    contents, so let's drop this line.
    And use "boot mode" instead of "boot type" to match
    https://eu.api.ovh.com/console/?section=%2Fdedicated%2Fserver&branch=v1#get-/dedicated/server/-serviceName-/specifications/hardware's
    return.

## Mandatory information


Y0Coss (approved):

JayBeeDe (approved):

sbraz:
@Y0Coss you can go ahead and merge this :)

Y0Coss:

PR #8461, opened 2025-10-01
KevS-SK2376-Update-New-SSL-management-for-WebHosting

Kazarus4:
Carroussel tabs with all new SSL screens added for : 

Add SSL LE
Add SSL comodo DV
Add SSL comodo EV
Add custom SSL
Delete SSL

this for all subs.

No proof needed because all of the content had be copied from SSL revamp PR (which was already proofed for all subs and still awaiting of merge at the end of migration.


Kazarus4:
Need proof by @benchbzh only on ES IT PL and PT.

PR #8527, opened 2025-10-15
[Fix] fixing wrong name for Milan DC

gbarideau:

- Fix

## Description

3-AZ Milan was wrongly named EU_WEST_MIL instead of EU_SOUTH_MIL

## Mandatory information


Y0Coss:

PR #8537, opened 2025-10-16
fix: s3-terraform: do not pin the ovh terraform provider to a specific version

rbeuque74:

- Fix

## Description

Terraform version of our provider was pinned to a specific version, which is old.
We should not encourage customers to pin a specific version, but use the latest one.

## Mandatory information


Y0Coss:
Thank you for the fix @rbeuque74

PR #8574, opened 2025-10-23
[AI Endpoints] - Replace removed models

MathieuBsqt:
Hello 👋

- Update of existing guide
- 
## Description

Riva model has been removed from AI Endpoints so we need to replace it by Whisper model in order to handle speech to text.
We will also replace the Mistral 8x22B by 8x7B version for same reason.

only `gb` version edited.

## Mandatory information


Y0Coss:
Thank you for the update @MathieuBsqt

PR #8584, opened 2025-10-27
lacp-resilience-scale-hg: mention affected commercial ranges

sbraz:
Servers with 2 interfaces are not affected.


Y0Coss (approved):

Y0Coss:
Thank you for the update @sbraz

PR #8605, opened 2025-10-30
[AI Endpoints] - Fix lib import

MathieuBsqt:
Hello 👋

- Fix

## Description
Langchain library has been updated so the import needs to change. 
However, I confirm we don't need to update the requirements.txt content.

## Mandatory information


Y0Coss:
Thank you for the fix @MathieuBsqt

PR #8616, opened 2025-10-31
Update Quantum documentation

wiseflat:

- New guide for Quantum QPUs
- Update of existing guide for Quantum emulators

## Description

Add new guide about QPUs

## Mandatory information

- This Pull Request shouldn't be merged before: 2025-11-06


Y0Coss:
You have a duplicated meta.yaml with the same ID and full_slug.
Overall, we do not recommend editing the ID and full_slug of an already-prodded guide. This would require redirections to be written.

wiseflat:
> You have a duplicated meta.yaml with the same ID and full_slug. Overall, we do not recommend editing the ID and full_slug of an already-prodded guide. This would require redirections to be written.

Is it better now ?

benchbzh:
Thank you for your contribution @wiseflat!

PR #8768, opened 2025-12-03
Restoring wrongfully deleted image on several guidess

Y0Coss:

- Fix

## Description

Yesterday a series of Pull Requests aimed at deleting unused images on all documentation were images.
A few of these images were actually still used in some guides.
This PR restores them.
FYI @DoStupidThingsFasterWithMoreEnergy


DoStupidThingsFasterWithMoreEnergy:
Hi @Y0Coss,

Thanks for the info.
After checking, I see that these images are used in HTML tags and not in Markdown tags.
I just pushed a pull request (#8769) because I also found errors in the HTML tags.

PR #8946, opened 2026-01-21
[Public Cloud DB] Fix the reorganization of PostgreSQL guides

thild42:
Create dedicated entry for PostgreSQL section as reusing the Common ones mess with the indexing.


Y0Coss:
Thank you for the fix @thild42

PR #8959, opened 2026-01-22
[Public Cloud DB] Rename Common and PostgreSQL guides

thild42:
A first batch of renaming


Y0Coss (approved):

Y0Coss:

PR #8966, opened 2026-01-23
[Public Cloud DB] Rename guides (second batch)

thild42:
Rename configuration guides


Y0Coss:

PR #8968, opened 2026-01-26
[Public Cloud DB] Update Valkey available versions and max connections formulaé

thild42:
Add version 8.1 as available.
Update the formula to showcase an example using the B3-8 instead of the DB1-7.


Y0Coss:

PR #8981, opened 2026-01-27
[Public Cloud DB] Reorganize and rename MySQL guides

thild42:
Normalize documentation categories.
Rename guide for clarity.


Y0Coss:

PR #8990, opened 2026-01-29
[Analytics] Rename common guides

thild42:
Rename common guide for clarity.
Reorganize common guides a bit.


Y0Coss:

PR #9003, opened 2026-02-02
[Public Cloud Analytics] Reorganize and rename Kafka guides

thild42:
Reindex kafka guides.
Rename guide for clarity.


Y0Coss:

PR #9033, opened 2026-02-09
[AI & Machine Learning] - AI Endpoints - Show  unauthenticated and authenticated request examples

MathieuBsqt:
Hello 👋

- Update of existing guie

## Description

A user has reported our examples were only showing unauthenticated mode, which is rate limited. It was indicated in the guide, but to make it better, we are going to show both approaches now:
- Unauthenticated (rate limited)
- Authenticated with AI Endpoints access key

## Mandatory information


Y0Coss:
Thank you for the update @MathieuBsqt

PR #9042, opened 2026-02-10
[AI & Machine Learning - AI Deploy] - Update Scaling Stategies guides

MathieuBsqt:
Hello 👋

- Update of existing guide(s)

## Description

We have prodded a new feature which allows our users to scale down their app to 0 replicas, allowing them to not pay anymore if their app has no load/usage. If the app has some load, it will reboot some replicas.

I've updated the guides to introduce this feature and also did a more general update of the scaling strategy guide.

## Mandatory information


MathieuBsqt:
I saw there were other PRs under review for Control Plane. Maybe it should be added in this new guide as well ?
Otherwise this PR is now ready to be prodded. We just need to update `updated` field to change date.

benchbzh:

PR #9153, opened 2026-03-06
VMware - Lifecycle Policy updates

P13K:

- Update of existing guide(s)

## Description

VMware - Lifecycle Policy updates: Add Granite lineup


Y0Coss:
Thank you for the update @P13K

PR #9194, opened 2026-03-18
Control Panel Templates - Public Cloud databases - Part 2

Jessica41:

- Optimization

## Description

Control Panel Templates - Public Cloud databases - Part 2

## Mandatory information

- [ ] OVHcloud integrated translation LLM 
- [ ] Systran
- [ ] Other tool (specify which tool was used)
- [ ] This Pull Request didn't require any translation.

- This Pull Request content should be replicated for the US OVHcloud documentation : YES/NO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ cp-extraction
 CLAUDE.md
 CLAUDE.local.md
 .mcp.json
+cp_template_inserter.py
+scripts/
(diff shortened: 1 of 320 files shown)


Y0Coss (changes requested):
Translated templates have been inserted but the guides are fully English. 
This needs to be updated so that only EN templates are inserted.

PR #9253, opened 2026-04-01
Integrations - Remove Integrations category and associated Prefect guides

MathieuBsqt:

- Deletion of guide(s) 

Guides are not mentioned in our documentation so no redirection is needed.

## Description

As discussed, we are going to remove the Integration category which is not filled enough. Associated guides will be also deleted as they re not used that much.

## Mandatory information

- I don't think these guides are on the US documentation.


Y0Coss:
Redirections PR created on Puppet repo, PR 8423. Now waiting for its merge.

Y0Coss:
Redirections are merged. Thank you for the update @MathieuBsqt

PR #9281, opened 2026-04-13
Bare Metal Cloud - lacp-netplan: drop main-private MAC heuristic

sbraz:

- Update of existing guide(s)
- Fix

## Description

The note suggesting that the main private interface could be identified as the one with the lowest MAC address value is no longer accurate and should not be relied upon. In the future, the fallback/main private interface will be exposed directly via the API.

This partially reverts c88844644c (ovh/docs#8833).

Ref: PUBM-52371

## Mandatory information

- [x] This Pull Request didn't require any translation.


benchbzh:
Thank you for your contribution @sbraz!

PR #9282, opened 2026-04-13
network_ipv6: disable randomized IPv6 IDs on Windows Server

sbraz:

- Update of existing guide(s)

## Description

Add a new step to the Windows Server 2019+ section instructing users to
run `Set-NetIPv6Protocol -RandomizeIdentifiers Disabled` from an elevated
PowerShell prompt. This is required for IPv6 connectivity on OVHcloud
infrastructure: it makes Windows build its IPv6 link-local addresses
from the adapter's MAC address (EUI-64) instead of using random values.
The change takes effect immediately and persists across reboots.

Without this setting, connectivity can silently break: TORs stop
forwarding inbound IPv6 traffic to the host after about 4 hours.

Future Windows Server installations performed through the OVHcloud
Control Panel / API will apply this setting automatically; the step is
documented here for existing installs and for customer-made manual
installs.

Ref: PUBM-52046

## Mandatory information

- [x] This Pull Request didn't require any translation.


DaMikeTheOne (approved):
Looks good to me

sbraz:
@Y0Coss can we merge this please? :)

Jessica41:

PR #9343, opened 2026-04-29
VPS - fix Nextcloud connectivity, pull errors, and persistence

afiand:

- Update of existing guide

## Description

**1. Networking & DNS Resolution**
* **Fix**: Introduced a dedicated `internal` network for the `db`, `redis`, and `app` services.
* **Reasons**: Resolved the error: `php_network_getaddresses: getaddrinfo for db failed`. Relying solely on the `external` proxy network caused Docker's internal service discovery to fail. A private bridge network ensures reliable communication between the application and its dependencies.

**2. Image Pull & Registry Compliance**
* **Fix**: Standardized the Redis image name to lowercase (`redis:7-alpine`).
* **Reasons**: Corrected the error: `invalid reference format: repository name must be lowercase`. Docker registry protocols require lowercase repository names; the previous uppercase "Redis" caused pull failures on standard Docker engines.

**3. Data Persistence & Integrity**
* **Fix**: Corrected the MariaDB volume path to `/var/lib/mysql` and added a persistent volume for the Nextcloud `/config` directory.
* **Reasons**: 
    * **MariaDB**: Linux filesystems are case-sensitive. 
    * **Config**: Without a volume for `/var/www/html/config`, users lose their `config.php` (and thus their entire setup) whenever the container is updated or recreated.

**4. Traefik Routing Optimization**
* **Fix**: Added the `traefik.docker.network=proxy` label to the `app` service.
* **Reasons**: Since the `app` container is now attached to two networks (`internal` and `proxy`), this label is required to explicitly tell Traefik which network to use for routing public web traffic, preventing potential 502 Gateway Errors.

## Mandatory information

### DIFF
--- a/pages/bare_metal_cloud/virtual_private_servers/install_nextcloud_on_vps_advanced/guide.en-gb.md
+++ b/pages/bare_metal_cloud/virtual_private_servers/install_nextcloud_on_vps_advanced/guide.en-gb.md
@@ -156,11 +156,15 @@ services:
       MYSQL_PASSWORD: ${DB_PASSWORD}
       MYSQL_ROOT_PASSWORD: ${DB_ROOT_PASSWORD}
     volumes:
-      - db_data:/var/lib/MySQL
+      - db_data:/var/lib/mysql
+    networks:
+      - internal
 
   Redis:
-    image: Redis:7-alpine
+    image: redis:7-alpine
     restart: unless-stopped
+    networks:
+      - internal
 
   app:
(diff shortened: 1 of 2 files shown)


Stephane-Bautista (commented):
Agree with that review

Y0Coss:
Now ported to the new repository supporting docs.ovhcloud.com via https://github.com/ovh/ovhcloud-docs/pull/23
Thank you for your contribution @afiand and sorry about the delay.

PR #9386, opened 2026-05-13
BMC Firmware Fix

SR-OVH-US:
Replicating an update made on the US documentation at the behest of the US Bare Metal team: "There should be no reference to versions being lower than '1.14' or anything else. Each motherboard will have its own version, so there is no unified version to compare with."

- Fix

## Description

Removing references to version 1.14, as requested by the US Bare Metal Team. "There should be no reference to versions being lower than '1.14' or anything else. Each motherboard will have its own versions, so there is no unified version to compare with."

## Mandatory information

- [ ] OVHcloud integrated translation LLM 
- [ ] Systran
- [ ] Other tool (specify which tool was used)
- [ ] This Pull Request didn't require any translation.

- This Pull Request content should be replicated for the US OVHcloud documentation: Completed

### DIFF
--- a/pages/bare_metal_cloud/dedicated_servers/bmc_fw_update/guide.en-gb.md
+++ b/pages/bare_metal_cloud/dedicated_servers/bmc_fw_update/guide.en-gb.md
@@ -50,9 +50,6 @@ sudo ipmitool mc info
 
 ![BMC firmware version output from ipmitool on Linux](images/ipmi_tool.png){.thumbnail}
 
-- If the firmware version is equal to or lower than 1.14, please contact our support team by creating a [support ticket via the OVHcloud Help Center](/links/support-contact) to request a firmware update. 
-- If the version is higher than 1.14, no action is required.
-
 ### On a Windows Server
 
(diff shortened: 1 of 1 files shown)


sbraz:
Hello @SR-OVH-US, we discussed this change with the Bare Metal Hardware team. The scope of this documentation is **only the TYAN S8056 board**. The change should be to clarify that in the summary/prerequisites. Maybe the path should be updated to match this too, @Y0Coss?

The references to a specific firmware version must remain.
This documentation complements the email which we sent to customers (PUBM-50687), in which we clearly stated:
>* **Affected motherboard :** TYAN S8056
>* **New BMC firmware version :** 2.00.0

If this documentation were to become more generic, we would probably need to update wording and screenshots with additional examples.

sbraz:
> If this documentation were to become more generic

For the record, I'm in favor of keeping separate pages for separate boards/BMCs, like we do for disk firmware upgrade guides.

bs-ovh:
Now here: https://github.com/ovh/ovhcloud-docs/pull/88

